PR #4, opened 2019-07-03
SFX-95: Setup testing environment

geoffcoutts:
- Setup for testing suite
  - Update dependencies
  - Update package scripts for running tests in browser and node
- Add sample module and test
- Update README with information for tests


README.md
```diff
@@ -30,6 +30,39 @@ To build an individual package in response to changes within the `src` directory
...
+
+To run tests for a specific web component package, navigate to its directory and use one of the following commands based on the desired testing flow:
+
+- To run the tests for a specific package once:
+```sh
+yarn test
+```
+- To run the tests for a specific package and watch the `src` and `test` directories to rerun the tests after any changes:
+```sh
+yarn tdd
```
DanielMoniz-GBI: I'm assuming `yarn tdd` is not intended to be used because people will use it when they are developing a specific component. If that's the case, maybe make it clear that `yarn test` works when run at the top level, but `yarn tdd` does not.

README.md
```diff
@@ -30,6 +30,39 @@ To build an individual package in response to changes within the `src` directory
...
+When starting a new component, create new `setup.ts` and `utils.ts` files in its new `tests` directory. These files will make the same files from the root `test-tools` directory available through imports and exports.
+
+Ex.
+- utils.ts
+```js
+export * from '../../../../../test-tools/utils';
+```
+- setup.ts
+```js
+import '../../../../../test-tools/setup';
```
DanielMoniz-GBI: Having to traverse 5 levels up is a red flag for me. We need to get some of these high-level folders (eg. `test-tools/`) on the path so we can import them directly.

This need not happen as part of this story/PR, but I wanted to point it out.

packages/web-components/@sfx/sayt/tsconfig.json
```diff
@@ -7,11 +7,14 @@
...
-      "node"
+      "node",
+      "mocha",
+      "chai",
+      "sinon"
     ],
     "module": "commonjs",
     "moduleResolution": "node",
-    "target": "ES5",
+    "target": "es2015",
```
DanielMoniz-GBI: I'm not too familiar with `tsconfig.json`, but is this intentional? Why are we no longer targeting ES5?

README.md
geoffcoutts: Clarified this in the README

packages/web-components/@sfx/sayt/tsconfig.json
geoffcoutts: As discussed, we'll be keeping this as `es2015` for now and will look to incorporate polyfills in the future.

README.md
geoffcoutts: As discussed, this will be reinvestigated later.

README.md
```diff
@@ -66,4 +104,4 @@ Storybook reference links:
 - [Writing Stories](https://storybook.js.org/docs/basics/writing-stories/)
 - [Using Knobs](https://github.com/storybookjs/storybook/tree/master/addons/knobs)
 
-\* One note about using knobs is that you have to use the **`text`** method for defining `array` or `object` properties for lit-element. 
+\* One note about using knobs is that you have to use the **`text`** method for defining `array` or `object` properties for lit-element.
```
jmbtutor: Did we not end up getting lit-html working in Storybook?

test-tools/setup.ts
```diff
@@ -0,0 +1,8 @@
+import { chai, sinon } from './utils';
```
jmbtutor: The files in this directory look the same as the one for the logic repo. We should consider consolidating these files into the devtools repo the future.

tsconfig.json
```diff
@@ -7,7 +7,9 @@
       "dom"
     ],
     "types": [
-      "node"
+      "node",
+      "mocha",
+      "chai"
```
jmbtutor: We aren't using `chai` in the global scope, so I don't think we need this.

packages/web-components/@sfx/sayt/test/index.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    const renderStub = stub(component, 'render').returns('component has rendered');
+    expect(component.render()).to.equal('component has rendered');
+    expect(renderStub).to.be.calledOnce;
+  });
+  it('should capitalize the placeholder term when rendering', () => {
+    const capitalizeStub = spy(component, 'capitalizePlaceholder');
+    const output = component.render();
+    expect(capitalizeStub).to.be.calledOnce;
+  });
+});
```
jmbtutor: I'm ignoring this file for now since it's full of dummy tests, but we should format the real tests better in the future.

README.md
DanielMoniz-GBI: I believe we did - what's the question here? In case it helps, here is some background.

This comment is clarifying that knobs are, frankly, annoying to use, and I would argue do not allow for client-presentable code. We'll be discussing this more in the coming sprints.

Additionally, given that Storybook is intended for use as how the client would mount our components and the logic layer, we cannot assume they are using LitElement.

packages/web-components/@sfx/sayt/test/index.test.ts
JoeyDeol: ^ I am also for better formatting in the future.

test-tools/setup.ts
geoffcoutts: Will make a follow up ticket to get that done in the future.

tsconfig.json
geoffcoutts: Have removed it from here and the top level `tsconfig.json`.

packages/web-components/@sfx/sayt/test/index.test.ts
geoffcoutts: Agreed.

PR #5, opened 2019-07-11
SFX-152 - Autocomplete sub-component, Base, Slots POC

ChantalDesrochers:
- Autocomplete component to receive data to populate into a list
-- sfx-list component created to be a generic list component
- Base class created
-- Slot POC contained within Base class

PR #6, opened 2019-07-12
SFX-200: Create Products wrapper

DanielMoniz-GBI:
Create a Products wrapper to wrap and display a number of individual Product tile components.


packages/web-components/@sfx/products/src/index.ts
```diff
@@ -0,0 +1,102 @@
...
+    `
+  }
+
+  createRenderRoot() {
+    return this;
+  }
+}
+
+@customElement('sfx-products')
+export class Products extends LitElement {
```
jmbtutor: I realize this is preliminary, but I would split this out into a separate file. `Product` as well, and have the index file simply re-export the individual module exports.

packages/web-components/@sfx/products/src/index.ts
```diff
@@ -0,0 +1,102 @@
...
+  @property({ type: Number }) maxItems = 12;
+  @property({ type: Array }) products = [];
```
jmbtutor: This should be typed. We might end up in an unexpected type situation because TS has to infer what it's an array _of_ just from an empty array.

packages/web-components/@sfx/products/src/index.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  constructor() {
+    super()
+    this.setProductsFromEvent = this.setProductsFromEvent.bind(this);
+  }
+
+  static get properties() {
```
jmbtutor: Same thing, these should be set in the property decorators.

packages/web-components/@sfx/products/src/index.ts
```diff
@@ -0,0 +1,102 @@
...
+    return {
+      maxItems: { reflect: true },
+    }
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.setUpEventListeners();
```
jmbtutor: We should have an equivalent in `disconnectedCallback()` to unregister event listeners to avoid memory leaks.

packages/web-components/@sfx/products/src/index.ts
```diff
@@ -0,0 +1,102 @@
...
+    this.products = event.detail.products;
+  }
+
+  getRenderableProducts() {
+    return this.products.slice(0, this.maxItems);
+  }
+
+  render() {
+    return html`
+      <style>
```
jmbtutor: Same as above, consider using the static styles getter.

packages/web-components/@sfx/products/stories/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  .add('Default', () => {
+    function sendSampleProducts() {
+      const productsEvent = new CustomEvent('sfx:provide-products', {
+        detail: {
+          products: sampleProducts,
+        }
+      });
+      window.dispatchEvent(productsEvent);
+    }
+    // @TODO Find other way of setting default products. Hacky.
```
jmbtutor: You can use [`window.customElements.whenDefined('sfx-products')`](https://stackoverflow.com/a/43434515).

packages/web-components/@sfx/products/test/index.test.ts
```diff
@@ -0,0 +1,95 @@
+import { expect, sinon, spy, stub } from './utils';
+import { Products } from '../src/index';
+
+describe('Products Component ', () => {
+  let component: any = {};
```
jmbtutor: No need to type-annotate nor provide a default value.

packages/web-components/@sfx/products/test/index.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    component = new Products();
+  });
+
+  describe('general', () => {
+    it('should default to have an empty array of products', () => {
+      expect(component.products).to.deep.equal([]);
+    });
+
+    it('should default to 12 maxItems', () => {
+      expect(component.maxItems).to.deep.equal(12);
```
jmbtutor: `.deep` is not necessary with primitives.

packages/web-components/@sfx/products/test/index.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    });
+  });
+
+  describe('properties', () => {
+    it('should reflect maxItems as a property', () => {
+      expect(Products.properties.maxItems.reflect).to.equal(true);
```
jmbtutor:
```suggestion
      expect(Products.properties.maxItems.reflect).to.be.true;
```

packages/web-components/@sfx/products/test/index.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+      expect(listenStub).to.be.called;
+    });
+  });
+
+  describe('setProductsFromEvent', () => {
+    it('should set the event products payload into the component', () => {
+      const products = [1, 2, 3];
+      const event = { detail: { products }};
+      expect(component.products).to.deep.equal([]);
```
jmbtutor: Is this assertion necessary? I would either set it explicitly to an empty array or assume that it's not going to be `products`.

packages/web-components/@sfx/products/test/index.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      component.setProductsFromEvent(event);
+
+      expect(component.products).to.equal(products);
+    });
+  });
+
+  describe('setUpEventListeners', () => {
+    it('should listen for a provide products event to set products', () => {
+      const listenStub = sinon.stub(window, 'addEventListener');
+      const callbackStub = sinon.stub(component, 'setProductsFromEvent');
```
jmbtutor: This function isn't being called, so it doesn't need to be stubbed.

packages/web-components/@sfx/products/test/index.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+      expect(listenStub).to.be.calledWith('sfx:provide-products', callbackStub);
+    });
+  });
+
+  describe('getRenderableProducts', () => {
+    it('should return zero products if there are none', () => {
+      const renderableProducts = component.getRenderableProducts();
+
+      expect(component.products.length).to.equal(0);
```
jmbtutor:
```suggestion
      expect(component.products).to.be.empty;
```

packages/web-components/@sfx/products/test/index.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      expect(renderableProducts.length).to.equal(0);
```
jmbtutor:
```suggestion
      expect(renderableProducts).to.be.empty;
```

packages/web-components/@sfx/products/test/index.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+      expect(renderableProducts).to.deep.equal([1, 2]);
+    });
+  });
+
+  describe('render', () => {
+    it('should return an object to be rendered', () => {
+      const result = component.render();
+
+      expect(result).to.be.an('object');
```
jmbtutor: We had a more correct test in #5.

packages/web-components/@sfx/products/README.md
```diff
@@ -0,0 +1,33 @@
+# SF-X Products Component
+
+## Functionality
+The `sfx-products` component acts as a wrapper around a series of individual
+`sfx-product` components. It uses a list of information about products
+and passes each item to a single product for rendering.
+
+### Received Events
+
+#### `sfx:provide-products`
```
jmbtutor:
```suggestion
#### `sfx::provide-products`
```

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * The sfx-products web component wraps and renders a number of
+ * sfx-product components. It wraps each sfx-product component in an
+ * additional wrapper for flexibility.
+ */
+@customElement('sfx-products')
+export default class Products extends Base {
+  @property({ type: Number, reflect: true }) maxItems = 12;
+  @property({ type: Array }) products: Product[] = [];
```
jmbtutor: These properties need doc comments.

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -0,0 +1,92 @@
...
+import { Base } from '@sfx/base';
+
+/**
+ * The sfx-products web component wraps and renders a number of
+ * sfx-product components. It wraps each sfx-product component in an
+ * additional wrapper for flexibility.
+ */
+@customElement('sfx-products')
+export default class Products extends Base {
+  @property({ type: Number, reflect: true }) maxItems = 12;
```
jmbtutor:
```suggestion
  @property({ type: Number, reflect: true }) maxItems: number = 12;
```

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -0,0 +1,92 @@
...
+    this.getRenderableProducts = this.getRenderableProducts.bind(this);
+  }
+
+  /**
+   * Registers event listeners.
+   */
+  connectedCallback() {
+    super.connectedCallback();
+
+    window.addEventListener('sfx:provide-products', this.setProductsFromEvent);
```
jmbtutor: Move this event name to a constant. I think we've also been using double colons.

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -0,0 +1,92 @@
+import {
+  customElement,
+  html,
+  property,
+  TemplateResult,
+} from 'lit-element';
+import Product from '@sfx/product/src/product';
```
jmbtutor: Does `Product` not get exported from `@sfx/product`?

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    window.removeEventListener('sfx:provide-products', this.setProductsFromEvent);
+  }
+
+  /**
+   * Sets the `products` property from the products in an event.
+   *
+   * @param event A custom event containing information about products.
+   */
+  setProductsFromEvent(event: CustomEvent) {
```
jmbtutor: The payload should be typed. Saying that this should be done when the events are consolidated in the new repo is also valid.

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -0,0 +1,92 @@
...
+  customElement,
+  html,
+  property,
+  TemplateResult,
+} from 'lit-element';
+import Product from '@sfx/product/src/product';
+import { Base } from '@sfx/base';
+
+/**
+ * The sfx-products web component wraps and renders a number of
```
jmbtutor:
```suggestion
 * The `sfx-products` web component wraps and renders a number of
```

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -0,0 +1,92 @@
...
+ * sfx-product components. It wraps each sfx-product component in an
```
jmbtutor:
```suggestion
 * `sfx-product` components. It wraps each `sfx-product` component in an
```

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -0,0 +1,92 @@
...
+   */
+  setProductsFromEvent(event: CustomEvent) {
+    this.products = event.detail.products;
+  }
+
+  /**
+   * Returns an array of products limited by the `maxItems` property.
+   */
+  getRenderableProducts(): Product[] {
+    return this.products.slice(0, this.maxItems);
```
jmbtutor: No action needed, but it's worth pointing out that we don't have a way to customize the start offset.

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -0,0 +1,92 @@
...
+        sfx-product {
+          width: 100%;
+          display: block;
+        }
+      </style>
+
+      ${this.getRenderableProducts().map(product => {
+        return html`
+          <div class="product-wrapper">
+            <sfx-product
```
jmbtutor: This is small enough that I think it can be put on one line.

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -0,0 +1,92 @@
...
+        .product-wrapper {
+          display: block;
+        }
+        sfx-product {
+          width: 100%;
+          display: block;
+        }
+      </style>
+
+      ${this.getRenderableProducts().map(product => {
```
jmbtutor: Semantically, I think this should generate a list.

packages/web-components/@sfx/products/stories/index.ts
```diff
@@ -0,0 +1,78 @@
+import { storiesOf } from '@storybook/html';
+import { withKnobs, number } from '@storybook/addon-knobs';
+
+import '../src/index.ts';
```
jmbtutor:
```suggestion
import '../src/index';
```

packages/web-components/@sfx/products/stories/index.ts
```diff
@@ -0,0 +1,78 @@
...
+import { ProductModel } from '../../product/src/product';
+
+const sampleProducts = [
```
jmbtutor: If you type this variable, you won't need to cast the values below.

```suggestion
const sampleProducts: ProductModel[] = [
```

packages/web-components/@sfx/products/stories/index.ts
```diff
@@ -0,0 +1,78 @@
...
+    function sendSampleProducts() {
+      const productsEvent = new CustomEvent('sfx:provide-products', {
+        detail: {
+          products: sampleProducts,
+        },
+      });
+      window.dispatchEvent(productsEvent);
+    }
+
+    window.customElements.whenDefined('sfx-products')
```
jmbtutor: Did we discover that this actually works?

packages/web-components/@sfx/products/test/index.test.ts
```diff
@@ -0,0 +1,97 @@
+import { TemplateResult } from 'lit-element';
+import { expect, sinon, spy, stub } from './utils';
+import Products from '../src/products';
+
+describe('Products Component ', () => {
+  let component;
+  beforeEach(() => {
```
jmbtutor: Add a blank line above.

packages/web-components/@sfx/products/test/index.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    component = new Products();
+  });
+
+  describe('general', () => {
```
jmbtutor: I would call this "properties" or whatever we ended up calling it in the other ones.

packages/web-components/@sfx/products/test/index.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    });
+
+    it('should default to 12 maxItems', () => {
+      expect(component.maxItems).to.equal(12);
+    });
+  });
+
+  describe('connectedCallback', () => {
+    it('should call super', () => {
+      const superConnected = stub(component.__proto__, 'connectedCallback');
```
jmbtutor: I would prefer to use `Object.getPrototypeOf(component)` instead of `__proto__` since the latter is just there for compatibility. The other tests go to the parent class directly and stub its prototype, but I think this is a more resilient approach.

packages/web-components/@sfx/products/test/index.test.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  describe('setProductsFromEvent', () => {
+    it('should set the event products payload into the component', () => {
+      const products = [1, 2, 3];
+      const event = { detail: { products }};
+      component.products = [];
+
+      component.setProductsFromEvent(event);
+
+      expect(component.products).to.equal(products);
```
jmbtutor: Do we care that it's the exact same? I think this should be `.deep.equal`.

packages/web-components/@sfx/products/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@sfx/products",
+  "version": "0.0.0",
+  "description": "",
+  "main": "dist/index.ts",
+  "types": "dist/index.d.ts",
```
jmbtutor: This is missing a `"modules": "src/index.ts"` declaration. Add it immediately above.

packages/web-components/@sfx/products/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@sfx/products",
+  "version": "0.0.0",
+  "description": "",
+  "main": "dist/index.ts",
```
jmbtutor:
```suggestion
  "main": "dist/index.js",
```

packages/web-components/@sfx/products/src/products.ts
geoffcoutts: We have been namespacing with `::`.
```suggestion
    window.addEventListener('sfx::provide-products', this.setProductsFromEvent);
```

packages/web-components/@sfx/products/src/products.ts
geoffcoutts: I'm under the impression that this component would be able to act as the main product grid on a PLP. Considering that, I'm not a fan of this property living in the component. I think the number of products being displayed should be limited by where the search call is made and then the product wrapper displays whatever number of products is provided to it. 

I believe the default number of products returned from a search request is 10. However, a client will need to be able to adjust that number before the search request is made, ideally in any sort of search driver they are using (either the one in the logic layer, or one that they build) as an option. If this property also lives in the product wrapper, then it basically just needs to mirror the number before the search call and is redundant. If for some reason the pre-search call property is ever higher than the products wrapper property, then some products will fail to display (creates a possible point of confusion for an implementer).

packages/web-components/@sfx/products/stories/index.ts
```diff
@@ -0,0 +1,78 @@
+import { storiesOf } from '@storybook/html';
+import { withKnobs, number } from '@storybook/addon-knobs';
+
+import '../src/index.ts';
+import { ProductModel } from '../../product/src/product';
```
geoffcoutts: Should be imported from the package
```suggestion
import { ProductModel } from '@sfx/product';
```

packages/web-components/@sfx/products/src/products.ts
jmbtutor: Thinking about it, I agree. This component should just display what it's given.

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -0,0 +1,96 @@
...
+  getRenderableProducts(): ProductModel[] {
+    return this.products.slice(0, this.maxItems);
+  }
+
+  render(): TemplateResult {
+    return html`
+      <style>
+        sfx-products {
+          display: flex;
+          flex-direction: row;
```
JoeyDeol: The default for flex-direction is row, you can remove this.

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -0,0 +1,96 @@
...
+          flex-wrap: wrap;
+        }
+        .product-wrapper {
```
JoeyDeol: Divs are display: block by default, you can remove this class selector and the display rule.

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -0,0 +1,96 @@
...
+          display: block;
+        }
+        sfx-product {
+          width: 100%;
```
JoeyDeol: Elements with display: block are width: 100% by default, you can remove this.

packages/web-components/@sfx/products/src/products.ts
DanielMoniz-GBI: @geoffcoutts @jmbtutor  Most of this makes sense to me, but it seems odd to not have the capability to limit the number of displayed items for a display component. It sounds like the problem here is that this forces the client to specify the information in both places. What if we default this value to `0` or `undefined`, and if it's that value, display every product it's given?

In that case, if the client specifies some number of search results to receive, they will not require specifying it here. So it wouldn't be redundant. Thoughts?

packages/web-components/@sfx/products/README.md
```diff
@@ -0,0 +1,33 @@
...
+### Received Events
+
+#### `sfx::provide-products`
+Upon receiving this event, the `sfx-products` component will accept
+the passed products data and update/add/remove any child product tiles.
+
+## Customization
+The SF-X Products component allows for configuring the quantity of displayed
+products. The following attributes are available:
+- `maxitems`: A number that limits the quantity of product tiles shown.
```
DanielMoniz-GBI: This should be removed now that the component no longer accepts this attribute. The sentence should really just be something like, "this component has no customization options."

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Registers event listeners.
+   */
+  connectedCallback() {
+    super.connectedCallback();
+
+
+    if (!this.getAttribute('role')) {
+      this.setAttribute('role', 'list');
```
DanielMoniz-GBI: What does the `role` actually do? Should `role` be a property?

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -0,0 +1,93 @@
...
+   * Binds relevant methods.
+   */
+  constructor() {
+    super();
+
+    this.setProductsFromEvent = this.setProductsFromEvent.bind(this);
+  }
+
+  /**
+   * Registers event listeners.
```
DanielMoniz-GBI: This comment should now be updated.

packages/web-components/@sfx/products/stories/index.ts
```diff
@@ -0,0 +1,75 @@
...
+    function sendSampleProducts() {
+      const productsEvent = new CustomEvent(PRODUCTS_EVENT, {
+        detail: {
+          products: sampleProducts,
+        },
+      });
+      window.dispatchEvent(productsEvent);
+    }
+
+    setTimeout(() => {
```
DanielMoniz-GBI: Can we use `customElements.whenDefined()` here?

packages/web-components/@sfx/products/src/products.ts
jmbtutor: `role` is the [ARIA role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles) for accessibility. Since this component is effectively a list and we aren't using a real list element like `ul`, setting `role` to `list` signals to the client that this element acts as a list.

The `HTMLElement` interface does not expose `role` as a property, so it must be set as an attribute.

packages/web-components/@sfx/products/stories/index.ts
jmbtutor: I think you originally used `whenDefined` and we decided through a thought experiment that we might run into a race condition since `whenDefined` says nothing about when a particular component has been upgraded. Granted, a timeout also introduces a race condition, but the risk is negligible given a sufficiently long timeout duration.

packages/web-components/@sfx/products/src/products.ts
ssbelle: @DanielMoniz-GBI Question: Should we declare this attribute earlier in the constructor ? I see examples in lit-element where we declare properties/attributes as static getters and then set them in the constructor?  https://lit-element.polymer-project.org/guide/properties#observed-attributes Just wondering

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -0,0 +1,94 @@
...
+
+        sfx-product {
+          display: block;
+        }
+      </style>
+
+      ${this.products.map(product => {
+        return html`
+          <div class="product-wrapper" role="listitem">
+            <sfx-product .product="${product}"></sfx-product>
```
ChantalDesrochers: I agree with @juliamlim 's suggestion to change the class name to 'product-tile-wrapper'.

packages/web-components/@sfx/products/src/products.ts
geoffcoutts: @DanielMoniz-GBI @jmbtutor I'm still of the opinion that ultimately it should live in whatever is driving the search. Wouldn't best practice for client implementations (using this product or not) be for them to specify their product count limit before a search occurs in order to minimize data retrieval and time? 
Ex. Searching for 30 products but then only limiting it to display the first 3 retrieves a lot of unnecessary data. 

What use cases do you foresee for using a max items value in the component?

PR #8, opened 2019-07-17
SFX 201: Product tile component

juliamlim:
Still have to work on unit tests


packages/web-components/@sfx/product/src/variants.ts
```diff
@@ -0,0 +1,31 @@
+import { LitElement, customElement, property, html } from 'lit-element';
+
+@customElement('sfx-product-variants')
+export class ProductVariants extends LitElement {
+  @property({ type: String }) type = '';
+  @property({ type: Array }) items = [];
+
+  listVariants() {
```
juliamlim: Forgot to finish this!

packages/web-components/@sfx/product/src/product.ts
```diff
@@ -0,0 +1,79 @@
+import { customElement, property, html } from 'lit-element';
+import Base from '../../base';
+
+import './variants';
+
+@customElement('sfx-product')
+export default class Product extends Base {
+  @property({ type: String }) display: 'full' | 'tile' = 'full';
```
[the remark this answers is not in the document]
juliamlim: I realized there isn't really a point to this property so I'm going to take it out

packages/web-components/@sfx/base/index.ts
```diff
@@ -0,0 +1,42 @@
+import { LitElement } from 'lit-element';
+
+export default class Base extends LitElement {
```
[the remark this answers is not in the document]
juliamlim: Yea this is an alternative way for slots to be rendered in the light dom but might need some refining. Right now it only allows for named slots and functions in the same way slots are used and written in the shadow dom.

packages/web-components/@sfx/product/src/product.ts
```diff
@@ -0,0 +1,79 @@
...
+  @property({ type: Object }) product: ProductModel = {};
+
+  urlWrap(url, children) {
+    return url ? html`<a href="${ url }">${children}</a>` : html`${children}`;
+  }
+
+  productNodes() {
+    const { product } = this;
+
+    const properties = Object.keys({
```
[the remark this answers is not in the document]
juliamlim: @DanielMoniz-GBI So one thing I noticed from doing this is that the dom doesn't update when you change the attribute or property. Something that I'm still working through but if you have time to 🍐that would be 🍇.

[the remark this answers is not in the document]
DanielMoniz-GBI: We hit the same issue with our existing slots solution. There may not be an easy solution to this without pulling in a library, and I'm not sure I like the library I've seen (vampire).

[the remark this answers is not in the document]
juliamlim: Maybe looking into how lit is handling the `attributeChangedCallback` and the rendering process might shed some light on how to resolve the issue, I'll check into that tomorrow.

DanielMoniz-GBI:
Is there a reason there are changes to `yarn.lock` in here? It looks like you've added a `package.json`, but that there are no dependencies in it. If there is no reason, let's isolate the changes to `yarn.lock` by leaving them for another PR (eg. next sprint).

packages/web-components/@sfx/product/stories/index.ts
```diff
@@ -0,0 +1,103 @@
...
+  promo: '25% off',
+  imageSrc: 'https://spudsmart.com/wp-content/uploads/2017/10/NZ-Bag-of-potatoes-cropped.png',
+  imageAlt: 'Good bag of potatoes'
+};
+
+storiesOf('Components|Product', module)
+  .addDecorator(withKnobs)
+  .add('Default', () => `
+    <sfx-product 
+      product="${text('Product Info', JSON.stringify({
```
DanielMoniz-GBI: Is it possible to use `.product` here to allow for passing a complex object directly (instead of using `JSON.stringify()`)?

packages/web-components/@sfx/product/src/variant.ts
```diff
@@ -0,0 +1,46 @@
...
+    this.setAttribute('role', 'listitem');
+    this.classList.add('product-variant');
+
+    switch (type) {
+      case 'color':
+        this.setAttribute('style', `background-color:${variant.color}`);
+        this.setAttribute('title', variant.text);
+        break;
+      case 'image':
+        this.setAttribute('style', `background-image:${variant.image};background-color:${variant.color}`);
```
jmbtutor: Should the image be a background image? The image is meaningful content if it's an image variant, so I think it should be put in an `` tag.

packages/web-components/@sfx/product/stories/index.ts
```diff
@@ -0,0 +1,103 @@
...
+              imageAlt: 'Mystery blue potatoes'
+            }, 
+          ] 
+        }
+      }))}"
+    ></sfx-product>
+  `, { 
+    notes: { 
+      markdown: `
+      # SF-X Product Component
```
DanielMoniz-GBI: I think everything in here should be indented.

packages/web-components/@sfx/product/src/variant.ts
```diff
@@ -0,0 +1,46 @@
...
+        this.setAttribute('title', variant.text);
+      case 'text':
+      default:
+        this.innerText = variant.text;
+    };
+
+    this.addEventListener('click', () => {
+      this.dispatchEvent(new CustomEvent('sfx::change_product_variant', {
```
jmbtutor: This is a reactive event, so it should be descriptive rather than imperative ("product\_variant\_change", for example). But take note of the comment above. As well, let's collect event names into a single file with exports so that it's easier to collect them into the new repo later on.

packages/web-components/@sfx/product/stories/index.ts
```diff
@@ -0,0 +1,103 @@
+import { storiesOf } from '@storybook/html';
+import { withKnobs, text } from '@storybook/addon-knobs';
+import '../src/index.ts';
```
jmbtutor: The extension is not needed.

```suggestion
import '../src/index';
```

packages/web-components/@sfx/product/stories/index.ts
```diff
@@ -0,0 +1,103 @@
...
+            { color: '#546', text: 'Blue', product: { 
+              imageSrc: 'http://startuptipsdaily.com/wp-content/uploads/2016/12/purple-and-blue-potato.jpg' },
+              imageAlt: 'Mystery blue potatoes'
+            }, 
+          ] 
+        }
+      }))}"
+    ></sfx-product>
+  `, { 
+    notes: {
```
DanielMoniz-GBI: I think we should ensure that the notes show up on all stories for this component.

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,199 @@
+import { TemplateResult, html } from 'lit-element';
+import { Base } from '@sfx/base';
+
+import { expect, stub, spy } from '../utils';
+
```
DanielMoniz-GBI: You can remove this newline.

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,231 @@
...
+  });
+
+  describe('updateVariant', () => {
+    it('should update the product property with new info', () => {
+      const event = {
+        detail: { name: 'Foo', price: 12.34 }
+      };
+      component.updateVariant(event);
+      expect(component.product.name).to.equal('Foo');
+      expect(component.product.price).to.equal(12.34);
```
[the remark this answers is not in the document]
jmbtutor: Is it a better test to assert the whole `component.product` instead? Set it to something known before calling the function and assert `component.product` with the entire expected value. I don't think it necessarily has to be a product that conforms to the real interface.

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,199 @@
...
+
+      component.disconnectedCallback();
+      expect(disconnectedCallback).to.have.been.called;
+    });
+
+    it('should remove eventListener from the window', () => {
+      const removeEventListener = spy(window, 'removeEventListener');
+
+      component.disconnectedCallback();
+      expect(removeEventListener).to.have.been.calledWith(
```
DanielMoniz-GBI: These tests should be updated to follow the convention we've been using: Arrange Act Assert (all separated by newlines). In this case, we just need a newline above the `expect` line here.

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,231 @@
...
+    });
+  });
+
+  describe('additionalInfo', () => {
+    it('should return an empty array if there are no additional properties', () => {
+      const result = component.additionalInfo();
+      expect(result).to.eql([]);
+    });
+
+    it('should return an array of TemplateResults there are additional properties', () => {
```
jmbtutor:
```suggestion
    it('should return an array of TemplateResults if there are additional properties', () => {
```

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,199 @@
...
+    });
+  });
+
+  describe('urlWrap', () => {
+    it('should return the children wrapped in an "a" tag', () => {
+      const result = component.urlWrap('string', html`<span>foo</span>`);
+      expect(result).to.be.an.instanceof(TemplateResult);
+    });
+
+    it('should return the children', () => {
```
DanielMoniz-GBI: Under what condition?

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,199 @@
...
+      component.render();
+      expect(additionalInfo).to.be.called;
+    });
+
+    it('should call the urlWrap function', () => {
+      const urlWrap = stub(component, 'urlWrap')
+      component.render();
+      expect(urlWrap).to.be.called;
+    });
+
```
DanielMoniz-GBI: Newline not needed.

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,199 @@
...
+  });
+})
+
+describe('Variant Component', () => {
```
DanielMoniz-GBI: I feel like these tests should be in a different file.

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,199 @@
...
+    component = new Variant();
+  });
+
+  describe('constructor', () => {
+    it('should extend the Base class', () => {
+      expect(component).to.be.an.instanceOf(Base);
+    });
+
+    describe('type property', () => {
+      it('should have a default value of "full"', () => {
```
DanielMoniz-GBI:
```suggestion
      it('should have a default value of "text"', () => {
```

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,199 @@
...
+        expect(component.type).to.equal('text');
+      });
+    });
+
+    describe('variant property', () => {
+      it('should have a default value of "full"', () => {
```
DanielMoniz-GBI: This test name should be updated.

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  describe('connectedCallback', () => {
+    it('should call its super connectedCallback', () => {
+      const connectedCallback = stub(Base.prototype, 'connectedCallback');
+
+      component.connectedCallback();
+      expect(connectedCallback).to.have.been.called;
+    });
+
+    it('should add an eventListener to the window', () => {
```
DanielMoniz-GBI: We might as well test which function gets registered as an event listener.

packages/web-components/@sfx/product/src/product.ts
```diff
@@ -0,0 +1,104 @@
...
+      </slot>
+      <slot name="price">
+        <p>${ price }</p>
+      </slot>
+      ${ this.additionalInfo() }
+    `;
+  }
+}
+
+export interface ProductModel {
```
jmbtutor: My guess why you declared the type of `this.product` as `any` is because of the possibility of additional properties? In that case, you're missing `[prop: string]: any;` in this interface. Whether you add it to `ProductModel` or you make a new interface that extends `ProductModel` to add that is up to you.

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,199 @@
...
+    it('should dispatch an event when component is clicked', () => {
+      const dispatchEvent = spy(component, 'dispatchEvent');
+
+      component.dispatchEvent(new Event('click'));
+      expect(dispatchEvent).to.have.been.called;
+    });
+  });
+
+  describe('render', () => {
+    it('should return a list element with text', () => {
```
DanielMoniz-GBI: I'm not sure this test is doing what its title suggests.

packages/web-components/@sfx/product/src/product.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export interface ProductVariantsModel {
+  type: 'color' | 'image' | 'text';
+  items: ProductVariantModel[];
+}
+
+export interface ProductVariantModel {
+  color?: string;
+  image?: string;
+  label?: string;
```
jmbtutor: It doesn't look like `label` is used anywhere. What's its purpose?

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,199 @@
...
+      const result = component.render();
+      expect(result).to.be.an.instanceof(TemplateResult);
+    });
+
+    it('should return a list element with a background color', () => {
```
DanielMoniz-GBI: Same as above.

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,199 @@
...
+      component.type = 'color';
+      component.variant = { text: 'foo', color: '#bed', product: {} };
+
+      const result = component.render();
+      expect(result).to.be.an.instanceof(TemplateResult);
+    });
+
+    it('should return a list element with a background image', () => {
```
DanielMoniz-GBI: Same as above.

packages/web-components/@sfx/product/stories/index.ts
jmbtutor: Not without lit-html working, which we've put aside for now.

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,231 @@
...
+    it('should return an array of TemplateResults there are additional properties', () => {
+      component.product = {
+        name: '',
+        salePrice: 0,
+        promo: '30% off'
+      };
+
+      const result = component.additionalInfo();
+
+      expect(result[1]).to.be.an.instanceOf(TemplateResult);
```
jmbtutor: I think we should assert the entire array. You can make multiple assertions on the length and the individual items, or make use of [Sinon matchers](https://sinonjs.org/releases/latest/matchers/).

packages/web-components/@sfx/product/src/product.ts
```diff
@@ -0,0 +1,104 @@
...
+    const product = e.detail;
+
+    this.product = {
+      ...this.product,
+      ...product
+    };
+  }
+
+  urlWrap(url: string, children: TemplateResult) {
+    return url ? html`<a href="${ url }">${children}</a>` : html`${children}`;
```
jmbtutor: If `children` is already a `TemplateResult`, wouldn't it be better to simply return it in the case where the `url` is falsy?

packages/web-components/@sfx/product/test/unit/product.test.ts
jmbtutor: Until we have structural testing in place, I don't think we can assert this unless it returns the argument itself.

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,231 @@
...
+    it('should add an eventListener to the window', () => {
+      const addEventListener = spy(component, 'addEventListener');
+
+      component.connectedCallback();
+      expect(addEventListener).to.have.been.called;
+    });
+
+    it('should set role to "listitem"', () => {
+      component.connectedCallback();
+      expect(component.getAttribute('role')).to.equal('listitem');
```
jmbtutor: This test should be stubbing `setAttribute` and asserting that it was called with the proper arguments instead.

packages/web-components/@sfx/product/stories/index.ts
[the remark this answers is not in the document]
juliamlim: Not using `Product`, forgot to remove that. Pulling in this because I just want to access when the window defines the components as custom elements.

packages/web-components/@sfx/product/stories/index.ts
juliamlim: Wondering if it's worth it to write something that will pull in the package's README.md file if it exists and if it doesn't or there's a string here it will be overwritten.

packages/web-components/@sfx/product/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@sfx/product",
+  "version": "0.0.0",
+  "description": "",
+  "main": "dist/index.js",
+  "module": "src/index.ts",
```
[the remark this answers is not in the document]
geoffcoutts: Should a `types` entry be added as well?

packages/web-components/@sfx/product/stories/index.ts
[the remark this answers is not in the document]
juliamlim: Ah just saw this comment, not at the moment no. I do think it's worth it to write something to allow us to do this however.

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,231 @@
...
+      });
+
+      it('should set innerText if the type is "text"', () => {
+        component.type = 'text';
+        component.connectedCallback();
+        expect(component.innerText).to.equal(component.variant.text);
+      });
+    });
+
+    it('should dispatch an event when component is clicked', () => {
```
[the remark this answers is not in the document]
juliamlim: This was something I wanted to test but I realized this should probably be left out until interaction testing is set up.

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,231 @@
...
+  describe('connectedCallback', () => {
+    it('should call its super connectedCallback', () => {
+      const connectedCallback = stub(Base.prototype, 'connectedCallback');
+
+      component.connectedCallback();
+      expect(connectedCallback).to.have.been.called;
+    });
+
+    it('should add an eventListener to the window', () => {
+      const addEventListener = spy(component, 'addEventListener');
```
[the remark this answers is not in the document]
juliamlim: Ah my bad they actually all should be component and not window

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,231 @@
...
+    it('should call the urlWrap function', () => {
+      const urlWrap = stub(component, 'urlWrap')
+      component.render();
+      expect(urlWrap).to.be.called;
+    });
+
+  });
+})
+
+describe('Variant Component', () => {
```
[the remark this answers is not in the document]
juliamlim: Was having an issue with importing { Base } twice when I had it separated in two different files.

packages/web-components/@sfx/product/test/unit/product.test.ts
juliamlim: re: https://github.com/groupby/sfx-view/pull/8/files#r307448566

packages/web-components/@sfx/product/test/unit/product.test.ts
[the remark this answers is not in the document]
juliamlim: Not sure what you mean here

packages/web-components/@sfx/product/stories/index.ts
```diff
@@ -0,0 +1,103 @@
...
+      yarn tdd
+      \`\`\`
+      `
+    } 
+}).add('Slotted Price', () => `
+  <sfx-product 
+    class="full"
+    product="${text('Product Info', JSON.stringify(product))}"
+  >
+    <h1 slot="title">Slotted Title</h1>
```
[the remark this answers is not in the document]
juliamlim: This format works in the separate base class I created but it's not being used atm so I can take this out for now.

packages/web-components/@sfx/product/src/variant.ts
```diff
@@ -0,0 +1,46 @@
...
+        bubbles: true,
+        detail: this.variant.product
+      }));
+    });
+  }
+
+  render() {
+    return html`<style>.product-variant { width: 15px; height: 15px; display: inline-block; margin: 2px; }</style>`;
```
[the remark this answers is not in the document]
juliamlim: Sorry wasn't sure if anyone replied to your comment in another PR. Because we're in the light dom the styles getter doesn't work.

packages/web-components/@sfx/product/src/variant.ts
```diff
@@ -0,0 +1,46 @@
...
+        break;
+      case 'image':
+        this.setAttribute('style', `background-image:${variant.image};background-color:${variant.color}`);
+        this.setAttribute('title', variant.text);
+      case 'text':
+      default:
+        this.innerText = variant.text;
+    };
+
+    this.addEventListener('click', () => {
```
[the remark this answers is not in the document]
juliamlim: The listener is targeting the single component and not the parent but is bubbling the event up to the parent object which is why it lives here.

packages/web-components/@sfx/product/src/variant.ts
```diff
@@ -0,0 +1,46 @@
...
+    this.classList.add('product-variant');
+
+    switch (type) {
+      case 'color':
+        this.setAttribute('style', `background-color:${variant.color}`);
+        this.setAttribute('title', variant.text);
+        break;
+      case 'image':
+        this.setAttribute('style', `background-image:${variant.image};background-color:${variant.color}`);
+        this.setAttribute('title', variant.text);
```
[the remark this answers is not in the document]
juliamlim: @jmbtutor for testing the style we'd need a package like chai-style, or should we wait for when we incorporate structural testing

packages/web-components/@sfx/product/src/product.ts
juliamlim: 🙏 ty needed this

packages/web-components/@sfx/product/src/product.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  render() {
+    const { name, price, variants, productUrl, imageSrc, imageAlt } = this.product;
+
+    return html`
+      <slot name="image">
+        ${ imageSrc ? html`<img src="${ imageSrc }" alt="${ imageAlt }" />`: '' }
+      </slot>
+      <slot name="variants">
```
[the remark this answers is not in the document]
juliamlim: There are some structural styles but this will most likely be stripped out when the global theme is done

packages/web-components/@sfx/product/src/product.ts
```diff
@@ -0,0 +1,103 @@
...
+      name: '',
+      price: 0,
+      variants: { type: 'text', items: [] },
+      productUrl: '',
+      imageAlt: '',
+      imageSrc: ''
+    };
+
+    return Object.keys(product).filter(p => !(p in properties)).map(p => html`
+      <span class="sfx-product-${ p }">${ product[p] }</span>
```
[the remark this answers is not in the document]
juliamlim: Yea we talked about emphasizing a flat object structure but there isn't a way to enforce it. Its the beginning of an idea but open to ideas!

packages/web-components/@sfx/product/src/product.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  urlWrap(url: string, children: TemplateResult) {
+    return url ? html`<a href="${ url }">${children}</a>` : html`${children}`;
+  }
+
+  additionalInfo() {
+    const { product } = this;
+
+    const properties: ProductModel = {
```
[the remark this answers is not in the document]
juliamlim: I wasn't able to get this to work, the only solution I found was https://stackoverflow.com/a/43922291 and it seemed like overkill to install a package for this one thing.

packages/web-components/@sfx/product/src/product.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  urlWrap(url: string, children: TemplateResult) {
+    return url ? html`<a href="${ url }">${children}</a>` : children;
+  }
+
+  additionalInfo() {
+    const { product } = this;
+
+    const properties: ProductModel = {
+      name: '',
```
DanielMoniz-GBI: If we're moving from a Product having a `name` to having a `title`, this should change too.

packages/web-components/@sfx/product/src/product.ts
```diff
@@ -0,0 +1,103 @@
...
+    return Object.keys(product).filter(p => !(p in properties)).map(p => html`
+      <span class="sfx-product-${ p }">${ product[p] }</span>
+    `);
+  }
+
+  render() {
+    const { name, price, variants, productUrl, imageSrc, imageAlt } = this.product;
+
+    return html`
+      <slot name="image">
```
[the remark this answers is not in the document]
DanielMoniz-GBI: We'll find out!

packages/web-components/@sfx/product/src/product.ts
```diff
@@ -0,0 +1,104 @@
...
+    return Object.keys(product).filter(p => !(p in properties)).map(p => html`
+      <span class="sfx-product-${ p }">${ product[p] }</span>
+    `);
+  }
+
+  render() {
+    const { name, price, variants, productUrl, imageSrc, imageAlt } = this.product;
+
+    return html`
+      <style>.product-variants { padding: 0 }</style>
```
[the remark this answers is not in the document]
DanielMoniz-GBI: Last I checked, this doesn't work with out current approach. I believe the static styles getter is intended for the Shadow DOM.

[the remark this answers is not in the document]
jmbtutor: Point taken. It looks like the [underlying functionality](https://github.com/Polymer/lit-element/blob/1385d948b6c0a991a1036e71d1f5c9e3858f3d6d/src/lit-element.ts#L177) relies on using the ShadyCSS polyfill.

packages/web-components/@sfx/product/src/product.ts
```diff
@@ -0,0 +1,104 @@
...
+  render() {
+    const { title, price, variants, productUrl, imageSrc, imageAlt } = this.product;
+
+    return html`
+      <style>.product-variants { padding: 0 }</style>
+      <slot name="image">
+        ${ imageSrc ? html`<img src="${ imageSrc }" alt="${ imageAlt }" />`: '' }
+      </slot>
+      <slot name="variants">
+        <ul class="product-variants">
```
jmbtutor: Should this be more specific to SF-X?

packages/web-components/@sfx/product/src/product.ts
```diff
@@ -0,0 +1,104 @@
...
+export interface ProductVariantsModel {
+  type: 'color' | 'image' | 'text';
+  items: ProductVariantModel[];
+}
+
+export interface ProductVariantModel {
+  color?: string;
+  image?: string;
+  text: string;
+  product: Partial<ProductModel>;
```
jmbtutor: `Partial` is not needed because all the properties are already optional.

packages/web-components/@sfx/product/src/product.ts
juliamlim: @geoffcoutts Leaving this until we decide on how elements are going to be targeted via `SFX-246`

packages/web-components/@sfx/product/src/variant.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  changeVariant() {
+    this.dispatchEvent(new CustomEvent('sfx::product_variant_change', {
+      bubbles: true,
+      detail: this.variant.product
+    }));
+  }
+
+  render() {
+    return html`<style>sfx-product-variant { width: 15px; height: 15px; display: inline-block; margin: 2px; }</style>`;
```
juliamlim: @geoffcoutts Marking this as structural styles for reference for `SFX-246`

packages/web-components/@sfx/product/test/unit/product.test.ts
```diff
@@ -0,0 +1,241 @@
...
+          product: {}
+        };
+      });
+
+      it('should set title and style if the type is "color"', () => {
+        component.type = 'color';
+
+        component.connectedCallback();
+
+        expect(component.style.backgroundColor).to.equal('rgb(187, 238, 221)');
```
jmbtutor: If we're going to be using this rgb value multiple times in the tests, I think it's worth it to put into a constant. Consider also replacing `#bed` with this, because both are valid colours.

PR #9, opened 2019-07-19
SFX 151 - Search Box Component

ChantalDesrochers:


packages/web-components/@sfx/search-box/stories/index.ts
```diff
@@ -0,0 +1,60 @@
...
+  console.log('event', e);
+});
+// end of event dispatch testing
+
+storiesOf('Components|Searchbox', module)
+  .addDecorator(withKnobs)
+  .add(
+    'Default - with search and clear button',
+    () => `
+  <sfx-search-box searchbutton clearbutton></sfx-search-box>
```
jmbtutor: Indent?

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+import SearchBox from '../../src/search-box';
+import { TemplateResult, html, LitElement } from 'lit-element';
+
+describe('SearchBox Component', () => {
+  let searchbox;
+  let windowDispatchEvent;
+
+  beforeEach(() => {
+    searchbox = new SearchBox();
+    windowDispatchEvent = spy(window, 'dispatchEvent');
```
jmbtutor: This should probably be a stub since you don't want it to actually dispatch the event.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+   */
+  describe('Constructor', () => {
+    it('should extend LitElement', () => {
+      expect(searchbox).to.be.an.instanceof(LitElement);
+    });
+  });
+
+  describe('connectCallback', () => {
+    it('should add an autocomplete_hover eventListener to the window', () => {
+      const windowAddEventListener = spy(window, 'addEventListener');
```
jmbtutor: This should also probably be a stub because you don't want it to actually listen for events.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+      expect(windowAddEventListener).to.have.been.calledWith(
+        SEARCHBOX_EVENT.AUTOCOMPLETE_HOVER,
+        searchbox.updateText
+      );
+    });
+  });
+
+  describe('disconnectCallback', () => {
+    it('should remove the autocomplete_hover eventListener from the window', () => {
+      const windowRemoveEventListener = spy(window, 'removeEventListener');
```
jmbtutor: Stub.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+      expect(windowRemoveEventListener).to.have.been.calledWith(
+        SEARCHBOX_EVENT.AUTOCOMPLETE_HOVER,
+        searchbox.updateText
+      );
+    });
+  });
+
+  describe('emitSearchEvent', () => {
+    it('should dispatch a search request event', () => {
+      searchbox.searchTerm = 'a';
```
jmbtutor: For ordering, I would put assignments after `const` and `let` declarations.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+      const searchRequestEvent = new CustomEvent(
+        SEARCHBOX_EVENT.SEARCH_REQUEST,
+        {
```
jmbtutor: I would put this on one line.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+  describe('emitAutocompleteRequestEvent', () => {
+    it('should dispatch an autocomplete event', () => {
+      const autocompleteRequestEvent = new CustomEvent(
+        'sfx::autocomplete_request',
+        { detail: 'b', bubbles: true }
+      );
+
+      searchbox.emitAutocompleteRequestEvent();
+
+      expect(windowDispatchEvent).to.have.been.calledWith(
```
jmbtutor: This should fit on one line.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+        autocompleteRequestEvent
+      );
+    });
+  });
+
+  describe('emitSearchBoxClearedEvent', () => {
+    it('should dispatch an emitSearchBoxClearedEvent', () => {
+      const searchboxxClearedEvent = new CustomEvent('sfx::search_box_cleared');
```
jmbtutor:
```suggestion
      const searchboxClearedEvent = new CustomEvent('sfx::search_box_cleared');
```

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+
+      searchbox.emitSearchBoxClearedEvent();
+
+      expect(windowDispatchEvent).to.have.been.calledWith(
```
jmbtutor: One line.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+        searchboxxClearedEvent
+      );
+    });
+  });
+
+  describe('updateText', () => {
+    it('should update the searchTerm property in response to data received', () => {
+      const detail = 'inputText';
+      stub(searchbox, 'getInputElement').returns(
```
jmbtutor: I'm not sure this is what you'd want to do, but perhaps this test might go away with the other suggested changes. We can discuss this.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+        `
+      );
+      searchbox.updateText({ detail });
+      expect(searchbox.searchTerm).to.equal(detail);
+    });
+  });
+
+  describe('handleKeyup', () => {
+    it('should remove the last letter of the searchTerm string if backpace is pressed', () => {
+      searchbox.searchTerm = 'hello';
```
jmbtutor: Ordering.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+
+      expect(searchbox.searchTerm).to.equal('hell');
+    });
+
+    it('should emit a searchBoxClearedEvent if backspace is pressed when there is one character remaining', () => {
+      const emitSearchBoxClearedEventSpy = spy(
+        searchbox,
+        'emitSearchBoxClearedEvent'
+      );
+      searchbox.searchTerm = 'h';
```
jmbtutor: Ordering

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+      );
+      searchbox.updateText({ detail });
+      expect(searchbox.searchTerm).to.equal(detail);
+    });
+  });
+
+  describe('handleKeyup', () => {
+    it('should remove the last letter of the searchTerm string if backpace is pressed', () => {
+      searchbox.searchTerm = 'hello';
+      const keyCode = 8;
```
jmbtutor: Since this is only used once, you can inline it.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+      expect(searchbox.searchTerm).to.equal('hell');
+    });
+
+    it('should emit a searchBoxClearedEvent if backspace is pressed when there is one character remaining', () => {
+      const emitSearchBoxClearedEventSpy = spy(
+        searchbox,
+        'emitSearchBoxClearedEvent'
+      );
+      searchbox.searchTerm = 'h';
+      const keyCode = 8;
```
jmbtutor: Since this is used once, you can inline it.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+
+      searchbox.handleKeyup({ keyCode });
+
+      expect(emitSearchBoxClearedEventSpy).to.have.been.called;
+    });
+
+    it('should invoke the emitSearchEvent function if enter is pressed and searchTerm property length is greater than 0', () => {
+      const emitSearchSpy = spy(searchbox, 'emitSearchEvent');
+      searchbox.searchTerm = 'hello';
```
jmbtutor: Ordering

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+      const keyCode = 13;
```
jmbtutor: Inline. I'll just say that this applies to all similar lines.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+
+      searchbox.handleKeyup({ keyCode });
+
+      expect(emitSearchSpy).to.have.been.calledWith('hello');
+    });
+
+    it('should invoke the updateSearchTerm function if word character is pressed', () => {
+      const updateSearchTermSpy = spy(searchbox, 'updateSearchTerm');
+      const keyCode = 68;
+      const target = {
```
jmbtutor: This can fit on one line.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+      searchbox.searchTerm = 'hello';
+      const keyCode = 13;
+
+      searchbox.handleKeyup({ keyCode });
+
+      expect(emitSearchSpy).to.have.been.calledWith('hello');
+    });
+
+    it('should invoke the updateSearchTerm function if word character is pressed', () => {
+      const updateSearchTermSpy = spy(searchbox, 'updateSearchTerm');
```
jmbtutor: You probably want this to be a stub.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+        value: 'dee'
+      };
+
+      searchbox.handleKeyup({ keyCode, target });
+
+      expect(updateSearchTermSpy).to.have.been.calledWith('dee');
+    });
+
+    it('should invoke the emitAutocompleteRequestEvent function if character key is pressed and searchTerm is 3 characters or more', () => {
+      stub(searchbox, 'updateSearchTerm');
```
jmbtutor: Ordering.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+      const emitAutocompleteSpy = spy(
+        searchbox,
+        'emitAutocompleteRequestEvent'
+      );
+      searchbox.searchTerm = 'darkkk';
+      const keyCode = 68;
+      const target = {
```
jmbtutor: This can fit on one line.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,239 @@
...
+    it('should set the search term property to the input value', () => {
+      searchbox.updateSearchTerm('catfood');
+
+      expect(searchbox.searchTerm).to.equal('catfood');
+    });
+  });
+
+  describe('clearSearch', () => {
+    it('should set the search term property to an empty string', () => {
+      stub(searchbox, 'getInputElement').returns(
```
jmbtutor: See comment for the similar block of code earlier.

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -0,0 +1,241 @@
...
+      this.updateText
+    );
+  }
+
+  /**
+   * Dispatches a search request event with the `searchTerm` property or string passed in.
+   * Invoked in response to user interactions: `enter` key or click on search button.
+   */
+  emitSearchEvent(query?: String) {
+    const term = query && typeof query === 'string' ? query : this.searchTerm;
```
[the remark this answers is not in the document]
DanielMoniz-GBI: Given that `this.searchTerm` is already stored, we should probably be going with that. If we come up with a use-case for being able to call `emitSearchEvent()` directly with a specific search term, we can go back to what Chantal has here.

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -0,0 +1,241 @@
...
+    const searchboxRequestEvent = new CustomEvent(
+      SEARCHBOX_EVENT.SEARCH_REQUEST,
+      {
+        detail: term,
+        bubbles: true
+      }
+    );
+    window.dispatchEvent(searchboxRequestEvent);
```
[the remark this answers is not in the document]
DanielMoniz-GBI: I think you're right. I feel like my old proof of concept dispatched from the `window`, but I can't remember why...Either way, we should be bubbling when possible so as to not create unexpected behaviour.

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -0,0 +1,241 @@
...
+    );
+    window.dispatchEvent(autocompleteRequestEvent);
+  }
+
+  /**
+   * Dispatches an event notifying that the input box has been cleared.
+   * Invoked in response to user input: clear button, remove all letters
+   * using keypresses.
+   */
+  emitSearchBoxClearedEvent() {
```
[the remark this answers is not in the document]
DanielMoniz-GBI: I think this was a request from me. It may be possible to handle it with what you described, but one way or another, we need to capture when the input is specifically cleared so we can act on it by closing the SAYT box. For now, we can probably just use Johann's suggestion of firing the "change" event that includes the current content of the search box.

For the future, I think the question is: are we okay with two events firing when the user clears the search box? The input would now have zero characters (one event), and the Clear button would have been clicked (another event).

packages/web-components/@sfx/search-box/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+Dispatched when a user clicks on the search button on hits `enter` within the search box. This event sends the search term value entered into the search box.
+
+#### `sfx::search_box_cleared`
+
+Dispatched when the value within the search box area is cleared. Occurs when a user clicks on the clear button, or when the final letter is removed from the search term.
+
+#### `sfx::autocomplete_request`
+
+Dispatched when the search term value within the search box contains more than 3 characters. This event sends the search term value entered into the search box.
```
JoeyDeol: Shouldn't there be a way for users to configure at what amount of characters the autocomplete request goes out? And by extension doesn't that mean that this event will trigger based on that configured number?

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -0,0 +1,241 @@
...
+    if (e.keyCode === 8) {
+      if (this.value.length === 1) {
+        this.emitSearchBoxClearedEvent();
+      }
+      this.value = this.value.slice(0, this.value.length - 1);
+    } else if (e.keyCode === 13 && this.value.length > 0) {
+      this.emitSearchEvent(this.value);
+    } else {
+      this.updatevalue((<HTMLInputElement>e.target).value);
+      if (this.value.length > 3) {
```
JoeyDeol: As I mentioned before, I think that the number of characters that an autocomplete request is triggered on should be configurable, so I think that this will have to change.

packages/web-components/@sfx/search-box/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+```sh
+yarn test
+```
+
+- To run the tests and watch the `src` and `test` directories to rerun the tests after any changes:
+
+```sh
+yarn tdd
+``
```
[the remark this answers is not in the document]
ChantalDesrochers: I've removed mention of tdd and only include the command to run the test once. Is this what you're requesting?

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -0,0 +1,241 @@
...
+    return html`
+    <input type="text" @mouseenter="${this.hoverExposed}" @click="${
+      this.clickExposed
+    }" id="searchInput" placeholder="${this.placeholderText}" @keyup="${
+      this.handleKeyup
+    }"></input>
+    ${
+      this.clearButton
+        ? html`
+            <button class="btn_clear" @click="${this.clearSearch}">
```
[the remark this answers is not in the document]
ChantalDesrochers: I've removed the classes for now, as based on previous conversations there will be a ticket that covers classes/styling.

packages/web-components/@sfx/search-box/README.md
ChantalDesrochers: Discussion with Johann lead to the conclusion that this should be handled by the component driver. I've removed all autocomplete events from this PR.

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -0,0 +1,205 @@
...
+  clickExposed() {
+    const searchBoxClickedEvent = new CustomEvent(SEARCHBOX_EVENT.SEARCHBOX_CLICK, { bubbles: true });
+    this.dispatchEvent(searchBoxClickedEvent);
+  }
+
+  /**
+   * Dispatches an event notifying that the search box input bar has been hovered.
+   * Invoked in response to a user hovering inside of the searchbox input.
+   */
+  hoverExposed() {
```
jmbtutor: What would the use case be for listening for this event? If we can't think of anything, perhaps we'll leave it out for now and put it in if someone asks for it later.

packages/web-components/@sfx/search-box/src/events.ts
```diff
@@ -0,0 +1,12 @@
+export const SEARCHBOX_EVENT = {
+  UPDATE_SEARCH_TERM: 'sfx::update_search_term',
+  SEARCH_REQUEST: 'sfx::search_request',
+  SEARCHBOX_CLEAR_CLICK: 'sfx::searchbox_clear_click',
+  SEARCHBOX_CLICK: 'sfx::searchbox_click_event',
+  SEARCHBOX_HOVER: 'sfx::searchbox_hover_event',
+  SEARCHBOX_CHANGE: 'sfx::on_searchbox_change'
```
jmbtutor: I would remove the `on_`. "On" should be for the listeners, but the event describes the actual action.

```suggestion
  SEARCHBOX_CHANGE: 'sfx::searchbox_change'
```

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -0,0 +1,205 @@
...
+    return html`
+    <input 
+      type="text" 
+      @mouseenter="${this.hoverExposed}" 
+      @click="${this.clickExposed}" 
+      data-sfx-ref="searchInput" 
+      placeholder="${this.placeholder}" 
+      @input="${this.handleChange}"
+      @keydown="${this.handleKeydown}">
+    </input>
```
jmbtutor: `<input>` is an empty element, so there's no closing tag.

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -0,0 +1,205 @@
...
+  render() {
+    return html`
+    <input 
+      type="text" 
+      @mouseenter="${this.hoverExposed}" 
+      @click="${this.clickExposed}" 
+      data-sfx-ref="searchInput" 
+      placeholder="${this.placeholder}" 
+      @input="${this.handleChange}"
+      @keydown="${this.handleKeydown}">
```
jmbtutor: For ordering, I would do `type`, `placeholder`, `data-sfx-ref`, then the event handlers in some order that makes sense.

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -0,0 +1,205 @@
...
+    </input>
+    ${
+      this.clearButton
+        ? html`<button @click="${this.clearSearch}">Clear</button>`
```
jmbtutor: No action needed for now, but do we want to consider setting the `type` of this button to `reset` in case the client puts this inside a `<form>`? If so, `clearSearch` would need to prevent default on the event.

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -0,0 +1,205 @@
...
+        : ''
+    }
+    ${
+      this.searchButton
+        ? html`<button @click="${this.emitSearchEvent}">Search</button>`
```
jmbtutor: Similarly, no action needed for now, but do we want to consider setting the `type` of this button to `submit` in case the client puts this inside a `<form>`? If so, `emitSearchEvent` would need to prevent default on the event.

packages/web-components/@sfx/search-box/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+#### `sfx::searchbox_hover_event`
+
+Dispatched when a user hovers over the search box.
+
+#### `sfx::search_request`
+
+Dispatched when a user clicks on the search button or hits `enter` within the search box. This event sends the search term value entered into the search box.
+
+#### `sfx::on_searchbox_change`
```
jmbtutor: See note about this in the events file.

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  /**
+   * Dispatches an `sfx::on_search_box_change` event, and updates the value
+   * property with the payload of the given event.
+   *
+   * Invoked in response to a change to the contents of the searchbox.
+   *
+   * @param e The KeyboardEvent object.
+   */
+  handleChange(e: KeyboardEvent) {
```
jmbtutor: Now that this is being bound to `oninput`, should we change this to `handleInput` instead?

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -0,0 +1,205 @@
...
+    this.updateSearchTermValue((e.target as HTMLInputElement).value);
+    this.dispatchEvent(
+      new CustomEvent(SEARCHBOX_EVENT.SEARCHBOX_CHANGE, {
+        detail: (e.target as HTMLInputElement).value,
+        bubbles: true
+      })
```
jmbtutor:
```suggestion
      });
```

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
jmbtutor: Yes, this is definitely not what you want. `getInputElement` only needs to return an empty object for the test not to fail.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,208 @@
...
+      const detail = 'inputText';
+      stub(searchbox, 'getInputElement').returns(
+        html`
+          <input type="text" data-sfx-ref="searchInput" placeholder="Type your search" />
+        `
+      );
+      searchbox.updateText({ detail });
+      expect(searchbox.value).to.equal(detail);
+    });
+  });
```
jmbtutor: Another test that you want is to make sure that the input element's `value` property is set to the search term. You'll want to stub `getInputElement` to return an empty object (or an object with a `value` property that is different than the expected value) and later expect that that object's `value` property contains the expected search term.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,208 @@
...
+  describe('handleKeydown', () => {
+    it('should invoke the emitSearchEvent function if enter is pressed and value property length is greater than 0', () => {
+      const emitSearchStub = stub(searchbox, 'emitSearchEvent');
+      searchbox.value = 'hello';
+
+      searchbox.handleKeydown({ keyCode: KEY_CODES.ENTER });
+
+      expect(emitSearchStub).to.have.been.called;
+    });
+  });
```
jmbtutor: I think you'd want to test the inverses as well: when a key other than "Enter" is pressed, and when the search term is empty.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,208 @@
...
+      const target = { value: 'dee' };
+
+      searchbox.handleChange({ target });
+
+      expect(updateSearchTermValueStub).to.have.been.calledWith('dee');
+    });
+
+    it('should dispatch a search box change event', () => {
+      const target = { value: 'dee' };
+      const searchboxChangeEvent = new CustomEvent('sfx::on_search_box_change', {
```
jmbtutor: Use the constant for the event name here.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,208 @@
...
+        html`
+          <input type="text" id="searchInput" placeholder="Type your search" />
+        `
+      );
+      
+      searchbox.clearSearch();
+
+      expect(emitSearchBoxSpy).to.have.been.called;
+    });
+  });
```
jmbtutor: This is missing the test for setting the value of the input box.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  describe('clickExposed', () => {
+    it('should dispatch a search box clicked event', () => {
+      const searchboxClickedEvent = new CustomEvent('sfx::search_box_cleared');
```
jmbtutor: Use the event name constant here.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,208 @@
...
+
+      searchbox.clickExposed();
+
+      expect(searchboxDispatchEvent).to.have.been.calledWith(searchboxClickedEvent);
+    });
+  });
+
+  describe('hoverExposed', () => {
+    it('should dispatch a search box hovered event', () => {
+      const hoverExposedEvent = new CustomEvent('sfx::search_hover_event');
```
jmbtutor: Use the event name constant here.

packages/web-components/@sfx/search-box/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+Dispatched when the value changes inside the search box input.
+
```
jmbtutor: I think you'd want to document the "update search" event as well.

packages/web-components/@sfx/search-box/src/search-box.ts
JoeyDeol: You're inside of a function call already, so the syntax is fine.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
JoeyDeol: Updated.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
JoeyDeol: I rolled this into the above test, I think that it helps solve the problem that we were seeing. Let me know what you think of the changes.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
JoeyDeol: Updated with similar code above, please review the approach.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
JoeyDeol: Rolled this into the first test, please review.

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  /**
+   * Invokes the emitSearchEvent function when a user presses enter and 
+   * the value property length is greater than 0.
+   *
+   * @param e The KeyboardEvent object.
+   */
+  handleKeydown(e: KeyboardEvent) {
+    if (e.keyCode === KEY_CODES.ENTER && this.value.length > 0) {
```
DanielMoniz-GBI: Apparently `.keyCode` is no longer recommended. You can use `.key` instead and check against `"Enter"`, which means we don't need to pull in a constant because it's self-documenting.

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  /**
+   * Invokes the emitSearchEvent function when a user presses enter and 
+   * the value property length is greater than 0.
+   *
+   * @param e The KeyboardEvent object.
+   */
+  handleKeydown(e: KeyboardEvent) {
+    if (e.key === "Enter" && this.value.length > 0) {
```
jmbtutor: I like double quotes, but we should have single quotes for consistency.

```suggestion
    if (e.key === 'Enter' && this.value.length > 0) {
```

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    describe('value property', () => {
+      it('should have default value of an empty string', () => {
+        expect(searchbox.value).to.equal('');
+      });
+    });
+
+    describe('searchButton property', () => {
+      it('should have default value of false', () => {
+        expect(searchbox.searchButton).to.equal(false);
```
jmbtutor:
```suggestion
        expect(searchbox.searchButton).to.be.false;
```

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      });
+    });
+
+    describe('clearButton property', () => {
+      it('should have default value of false', () => {
+        expect(searchbox.clearButton).to.equal(false);
```
jmbtutor:
```suggestion
        expect(searchbox.clearButton).to.be.false;
```

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,193 @@
...
+
+      expect(searchboxDispatchEvent).to.have.been.calledWith(searchboxClearedEvent);
+    });
+  });
+
+  describe('updateText', () => {
+    it('should update the value property of the input element in response to data received', () => {
+      const detail = 'inputText';
+      const inputObject = { value: 'Placeholder text' };
+      const inputEvent = new CustomEvent(SEARCHBOX_EVENT.UPDATE_SEARCH_TERM, { detail: detail });
```
jmbtutor:
```suggestion
      const inputEvent = new CustomEvent(SEARCHBOX_EVENT.UPDATE_SEARCH_TERM, { detail });
```

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
jmbtutor: I'm split on whether it should be a separate test or if this is okay, but if you're going to keep this a single test, then the second assertion should be directly asserting that `inputObject.value` is the expected value and not doing a roundabout assertion.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      searchbox.handleKeydown({ key: "Enter" });
+
+      expect(emitSearchStub).to.have.been.called;
+    });
+  });
+
+  describe('handleInput', () => {
+    it('should invoke the updateSearchTermValue function', () => {
+      const updateSearchTermValueStub = stub(searchbox, 'updateSearchTermValue');
+      const target = { value: 'dee' };
```
jmbtutor: `'dee'` should be factored out into a constant (and used in the assertion as well), but the `target` object can be inlined because it's not used more than once.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,193 @@
...
+
+      searchbox.handleInput({ target });
+
+      expect(searchboxDispatchEvent).to.have.been.calledWith(searchboxChangeEvent);
+    });
+  });
+
+  describe('updateSearchTermValue', () => {
+    it('should set the search term property to the input value', () => {
+      searchbox.updateSearchTermValue('catfood');
```
jmbtutor: `'catfood`' should be put in a constant.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,193 @@
...
+  describe('clearSearch', () => {
+    it('should set the search term property to an empty string', () => {
+      const inputObject = { value: 'Placeholder text' };
+      stub(searchbox, 'getInputElement').returns(inputObject);
+      stub(searchbox, 'emitSearchBoxClearClick');
+
+      searchbox.clearSearch();
+
+      expect(searchbox.value).to.equal('');
+      expect(searchbox.value).to.equal(inputObject.value);
```
jmbtutor: If you want to assert that `inputObject.value` should be an empty string, then simply assert that.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -0,0 +1,193 @@
...
+    });
+
+    it('should invoke the emitSearchBoxClearClick', () => {
+      const emitSearchBoxSpy = spy(searchbox, 'emitSearchBoxClearClick');
+      stub(searchbox, 'getInputElement').returns(
+        html`
```
jmbtutor: There's no need to return a `TemplateResult` here; it wouldn't work properly anyway. If you don't care about what it returns (you don't in this case), you can simply return an empty object.

PR #11, opened 2019-07-23
SFX-192: Create CircleCI Workflow

geoffcoutts:
Create jobs and workflow for building and testing of the repo.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,9 @@ defaults:
     command: |
       [[ ! -r ../build.tar ]] || tar xvf ../build.tar
 
+  git_submodule_install: &git_submodule_install
+    name: Install git submodules
+    command: git submodule update --init
   yarn_install: &yarn_install
     name: Install
     command: yarn
@@ -60,6 +63,7 @@ jobs:
       - run:
           name: Dependencies
           command: sudo apt install -y rsync
+      - run: *git_submodule_install
       - run: *yarn_install
       - save_cache:
           <<: *yarn_cache
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,9 +97,9 @@ workflows:
   build_test:
     jobs:
       - generate_cache_key_file:
-          filters:
-            branches:
-              ignore: /release\/.*/
+          # filters:
+          #   branches:
+          #     ignore: /release\/.*/
 
       - build:
           name: build_base
@@ -109,6 +109,10 @@ workflows:
           name: build_ui
           requires:
             - build_base
+      - build:
+          name: build_autocomplete
+          requires:
+            - build_ui
       - build_final:
           requires:
             - build_base
@@ -119,4 +123,7 @@ workflows:
           <<: *requires_build
       - unit test:
           name: unit_test_ui
+          <<: *requires_build
+      - unit test:
+          name: unit_test_autocomplete
           <<: *requires_build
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,188 @@
...
+  image: &image
+    docker:
+      - image: circleci/node:11.6.0-browsers
+  yarn_cache: &yarn_cache
+    key: yarn_cache-v{{ .Environment.CI_CACHE_VERSION }}-{{ .Branch }}-{{ checksum "yarn.lock" }}
+  build_final_cache_partial: &build_final_cache_partial
+    key: build_cache-v{{ .Environment.CI_CACHE_VERSION }}-{{ .Branch }}-{{ .Revision }}-{{ checksum "../workflow-start" }}-all-
+  archive_build: &archive_build
+    name: Archive build
+    command: tar cvf ../build.tar packages/web-components/@sfx/*/{dist,esnext}
```
jmbtutor: This is fine for now, but this will have to change when we add the wrappers (or anything that's not under `web-components`). If the structure is always going to be `@sfx` under a single subdirectory under `packages`, we should just change this to `packages/*/@sfx/*/{dist,esnext}` now.

.circleci/config.yml
```diff
@@ -0,0 +1,188 @@
...
+      - build:
+          name: build_sayt
+          <<: *requires_build_ui
+      - build_final:
+          requires:
+            - build_base
+            - build_ui
+            - build_autocomplete
+            - build_sayt
+
```
DanielMoniz-GBI: Please drop one of these newlines.

.circleci/config.yml
```diff
@@ -0,0 +1,188 @@
...
+      - restore_cache: *build_final_cache_partial
+      - run: *extract_build_archive
+      - run: *yarn_install
+      - run:
+          name: Test
+          command: |
+            package="${CIRCLE_JOB#unit_test_}"
+            package="${package//_/-}"
+
+            cd "packages/web-components/@sfx/${package}"
```
jmbtutor: My assumption is that we'll have unique package names, so if the assumption that I made in the defaults section is true, we can change this to `cd packages/*/"@sfx/${package}"` so that when we add the wrappers, we don't have to remember to change this (or worse, not realize that we don't have to make a new job).

.circleci/config.yml
```diff
@@ -0,0 +1,188 @@
...
+          name: build_sayt
+          <<: *requires_build_ui
+      - build_final:
+          requires:
+            - build_base
+            - build_ui
+            - build_autocomplete
+            - build_sayt
+
+
```
jmbtutor: Just one blank line is enough.

.circleci/config.yml
```diff
@@ -0,0 +1,188 @@
...
+            cd "packages/web-components/@sfx/${package}"
+            yarn build
+      - run: *archive_build
+      - save_cache:
+          key: build_cache-v{{ .Environment.CI_CACHE_VERSION }}-{{ .Branch }}-{{ .Revision }}-{{ checksum "../workflow-start" }}-{{ .Environment.CIRCLE_JOB }}-{{ epoch }}
+          paths:
+            - ~/build.tar
+
+  build_final:
```
jmbtutor: Add a doc comment to this.

.circleci/config.yml
geoffcoutts: Beat you to it ^

package.json
```diff
@@ -5,6 +5,8 @@
   "private": true,
   "workspaces": [
     ".storybook/*",
+    "packages/web-components/@sfx/base",
+    "packages/web-components/@sfx/ui",
     "packages/web-components/@sfx/*"
```
jmbtutor: With the assumption above, do we want to change this to `packages/*/@sfx/*`?

PR #14, opened 2019-08-01
SFX-245: Integrate show/hide functionality for SAYT

DanielMoniz-GBI:
Ensures SAYT is shown/hidden when required, rather than simply listening for show/hide events.

From this PR:
* Update SAYT to accept a `searchbox` attribute
* Update SAYT to only show/hide itself when the correct `searchbox` attribute is provided (or none at all).
* Add optional `Close` link to SAYT
  * Allow for `Close` link text to be customizable (eg. `X`)
* Updated all Search events to include the ID of the search box
  * Updated Search tests accordingly
* Update Search event-based tests for validity (they were not testing what we thought they were)
* Update SAYT Storybook stories to output the code required to produce the given result
  * Added `html-entities` as a requirement
* Listen on `autocomplete-receive-results` event to show SAYT (ie. automatically show SAYT once a subcomponent has results to show)
* Unit tests for all new functions in SAYT component
* Minor fixes to Search Storybook stories

Caveats/TO DO:
* No story added for Search + SAYT because they do not currently interact directly. They require the Logic Layer (SAYT Driver) to interact. In the meantime, there are stories that use `<input>` boxes.
* styles for Storybook code blocks are `<style>` tags inside of a function, rather than being an imported stylesheet (had trouble getting this to work).
* Syntax highlighting needed for display code (XML)
* Code output is still needed for all other Storybook stories (eg. for Search, Autocomplete, and Base)

### DIFF
--- a/.storybook/common.ts
+++ b/.storybook/common.ts
@@ -11,6 +11,6 @@ export function getDisplayCode(code: string): string {
         }
       </style>
       <h3>The code</h3>
-      <pre class="code-output"><code>${entities.encode(code)}</code></pre>
+      <pre class="code-output"><code>${ entities.encode(code) }</code></pre>
     `
   }
--- a/.storybook/common.ts
+++ b/.storybook/common.ts
@@ -1,0 +1,16 @@
+import { XmlEntities } from 'html-entities';
+
+const entities = new XmlEntities();
+
+export function getDisplayCode(code: string): string {
+    return `
+      <style>
+        pre.code-output {
+          padding: 15px;
+          background-color: #EEEEEE;
+        }
+      </style>
+      <h3>The code</h3>
+      <pre class="code-output"><code>${entities.encode(code)}</code></pre>
+    `
+  }
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -66,6 +66,7 @@ export default class Sayt extends LitElement {
     super.disconnectedCallback();
 
     window.removeEventListener(SAYT_EVENT.SAYT_SHOW, this.showCorrectSayt);
+    window.removeEventListener(AUTOCOMPLETE_RECEIVED_RESULTS_EVENT, this.showCorrectSayt);
     window.removeEventListener(SAYT_EVENT.SAYT_HIDE, this.hideCorrectSayt);
     window.removeEventListener('click', this.processClick);
     window.removeEventListener('keypress', this.processKeyPress);
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -185,9 +185,9 @@ export default class Sayt extends LitElement {
   render() {
     return html`
       ${ this.showCloseButton ?
-        html`<a href="#" aria-label="Close" .onclick=${ this.clickCloseSayt }>
+        html`<button aria-label="Close" .onclick=${ this.clickCloseSayt }>
           ${ this.closeText }
-        </a>`
+        </button>`
         : ``
       }
       ${ this.hideAutocomplete ? '' : html`<sfx-autocomplete></sfx-autocomplete>` }
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -165,7 +165,7 @@ export default class Sayt extends LitElement {
    */
   nodeInSearchBar(node: Node): boolean {
     if (!this.searchbox) return false;
-    const searchBar = document.querySelector('#' + this.searchbox);
+    const searchBar = document.getElementById(this.searchbox);
     return !!searchBar && searchBar.contains(node);
   }
 
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -95,7 +95,7 @@ export default class Sayt extends LitElement {
 
   /**
    * Makes SAYT visible if the event refers to the correct SAYT component.
-   * See isCorrectSayt().
+   * @see [[isCorrectSayt]]
    *
    * @param event An event that can contain a searchbox ID.
    */
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -19,6 +19,10 @@ export default class Sayt extends LitElement {
    * Stores the ID of the relevant search element.
    */
   @property({ type: String, reflect: true }) searchbar = '';
+  /**
+   * Customizes the text in the close button.
+   */
+  @property({ type: String }) closeText = 'Close';
 
   /**
    * Calls superclass constructor and bind methods.
@@ -174,7 +178,7 @@ export default class Sayt extends LitElement {
   render() {
     return html`
       ${ this.hideAutocomplete ? '' : html`<sfx-autocomplete></sfx-autocomplete>` }
-      <a href="#" .onclick=${this.clickCloseSayt}>Close</a>
+      <a href="#" aria-label="Close" .onclick=${this.clickCloseSayt}>${this.closeText}</a>
     `;
   }
 }
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -82,6 +82,11 @@ export default class Sayt extends LitElement {
     this.visible = true;
   }
 
+  /**
+   * Makes SAYT visible if the event refers to the correct SAYT component.
+   *
+   * @param event An event that can contain a searchbar ID.
+   */
   showCorrectSayt(event: CustomEvent) {
     if (this.isCorrectSayt(event)) {
       this.showSayt();
@@ -95,12 +100,24 @@ export default class Sayt extends LitElement {
     this.visible = false;
   }
 
+  /**
+   * Hides SAYT if the event refers to the correct SAYT component.
+   *
+   * @param event An event that can contain a searchbar ID.
+   */
   hideCorrectSayt(event: CustomEvent) {
     if (this.isCorrectSayt(event)) {
       this.hideSayt();
     }
   }
 
+  /**
+   * Determines whether an event refers to the correct SAYT. This is true if
+   * either a matching `searchbar` ID is specified, or if no `searchbar` ID
+   * is specified at all.
+   *
+   * @param event An event that can contain a searchbar ID for comparison.
+   */
   isCorrectSayt(event: CustomEvent) {
     const searchbar = event.detail.searchbar;
     return searchbar === this.searchbar || searchbar === undefined;
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -1,5 +1,6 @@
 import { LitElement, customElement, html, property, PropertyValues } from 'lit-element';
 import { SAYT_EVENT } from './events';
+import { AUTOCOMPLETE_RECEIVED_RESULTS_EVENT } from '../../autocomplete/src/events';
 
 /**
  * The `sfx-sayt` component is responsible for displaying and hiding the
@@ -52,6 +53,7 @@ export default class Sayt extends LitElement {
     super.connectedCallback();
 
     window.addEventListener(SAYT_EVENT.SAYT_SHOW, this.showCorrectSayt);
+    window.addEventListener(AUTOCOMPLETE_RECEIVED_RESULTS_EVENT, this.showCorrectSayt);
     window.addEventListener(SAYT_EVENT.SAYT_HIDE, this.hideCorrectSayt);
     window.addEventListener('click', this.processClick);
     window.addEventListener('keypress', this.processKeyPress);
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -149,7 +149,7 @@ export default class Sayt extends LitElement {
   }
 
   /**
-   * Handles hiding SAYT on click on a close link/button (or other event).
+   * Handles hiding SAYT on click of a close link/button (or other event).
    *
    * @param event An event with a default action to be prevented.
    */
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -42,7 +42,7 @@ export default class Sayt extends LitElement {
   connectedCallback() {
     super.connectedCallback();
 
-    window.addEventListener(SAYT_EVENT.SAYT_SHOW, this.showSayt);
+    window.addEventListener(SAYT_EVENT.SAYT_SHOW, this.showCorrectSayt);
     window.addEventListener(SAYT_EVENT.SAYT_HIDE, this.hideCorrectSayt);
     window.addEventListener('click', this.processClick);
     window.addEventListener('keypress', this.processKeyPress);
@@ -54,7 +54,7 @@ export default class Sayt extends LitElement {
   disconnectedCallback() {
     super.disconnectedCallback();
 
-    window.removeEventListener(SAYT_EVENT.SAYT_SHOW, this.showSayt);
+    window.removeEventListener(SAYT_EVENT.SAYT_SHOW, this.showCorrectSayt);
     window.removeEventListener(SAYT_EVENT.SAYT_HIDE, this.hideCorrectSayt);
     window.removeEventListener('click', this.processClick);
     window.removeEventListener('keypress', this.processKeyPress);
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -43,7 +43,6 @@ export default class Sayt extends LitElement {
     this.hideCorrectSayt = this.hideCorrectSayt.bind(this);
     this.showCorrectSayt = this.showCorrectSayt.bind(this);
     this.isCorrectSayt = this.isCorrectSayt.bind(this);
-    this.clickCloseSayt = this.clickCloseSayt.bind(this);
   }
 
   /**
@@ -185,7 +184,7 @@ export default class Sayt extends LitElement {
   render() {
     return html`
       ${ this.showCloseButton ?
-        html`<button aria-label="Close" .onclick=${ this.clickCloseSayt }>
+        html`<button aria-label="Close" @click=${ this.clickCloseSayt }>
           ${ this.closeText }
         </button>`
         : ``
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -131,7 +131,7 @@ export default class Sayt extends LitElement {
    *
    * @param event An event that can contain a searchbox ID for comparison.
    */
-  isCorrectSayt(event: CustomEvent) {
+  isCorrectSayt(event: CustomEvent): boolean {
     const searchbox = event.detail && event.detail.searchbox;
     return this.searchbox === undefined ||
       searchbox === this.searchbox ||
@@ -165,7 +165,7 @@ export default class Sayt extends LitElement {
    * 
    * @param node The node to check for containment.
    */
-  nodeInSearchBar(node: Node) {
+  nodeInSearchBar(node: Node): boolean {
     if (!this.searchbox) return false;
     const searchBar = document.querySelector('#' + this.searchbox);
     return !!searchBar && searchBar.contains(node);
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -34,6 +34,7 @@ export default class Sayt extends LitElement {
     this.hideCorrectSayt = this.hideCorrectSayt.bind(this);
     this.showCorrectSayt = this.showCorrectSayt.bind(this);
     this.isCorrectSayt = this.isCorrectSayt.bind(this);
+    this.clickCloseSayt = this.clickCloseSayt.bind(this);
   }
 
   /**
@@ -138,6 +139,16 @@ export default class Sayt extends LitElement {
   }
 
   /**
+   * Handles hiding SAYT on click on a close link/button (or other event).
+   *
+   * @param event An event with a default action to be prevented.
+   */
+  clickCloseSayt(event: Event) {
+    event.preventDefault();
+    this.hideSayt();
+  }
+
+  /**
    * Checks whether a given node is inside of SAYT's identified search bar.
    * 
    * @param node The node to check for containment.
@@ -163,6 +174,7 @@ export default class Sayt extends LitElement {
   render() {
     return html`
       ${ this.hideAutocomplete ? '' : html`<sfx-autocomplete></sfx-autocomplete>` }
+      <a href="#" .onclick=${this.clickCloseSayt}>Close</a>
     `;
   }
 }
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -119,7 +119,7 @@ export default class Sayt extends LitElement {
    * @param event An event that can contain a searchbar ID for comparison.
    */
   isCorrectSayt(event: CustomEvent) {
-    const searchbar = event.detail.searchbar;
+    const searchbar = event.detail && event.detail.searchbar;
     return this.searchbar === undefined ||
       searchbar === this.searchbar ||
       searchbar === undefined;
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -158,7 +158,7 @@ export default class Sayt extends LitElement {
   }
 
   /**
-   * Checks whether a given node is inside of SAYT's identified search bar.
+   * Checks whether a given node is inside of SAYT's identified search box.
    * 
    * @param node The node to check for containment.
    */
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -23,6 +23,10 @@ export default class Sayt extends LitElement {
    * Customizes the text in the close button.
    */
   @property({ type: String }) closeText = 'Close';
+  /**
+   * Shows a button to allow for closing SAYT manually.
+   */
+  @property({ type: Boolean }) showCloseButton = false;
 
   /**
    * Calls superclass constructor and bind methods.
@@ -177,8 +181,13 @@ export default class Sayt extends LitElement {
    */
   render() {
     return html`
+      ${ this.showCloseButton ?
+        html`<a href="#" aria-label="Close" .onclick=${this.clickCloseSayt}>
+          ${this.closeText}
+        </a>`
+        : ``
+      }
       ${ this.hideAutocomplete ? '' : html`<sfx-autocomplete></sfx-autocomplete>` }
-      <a href="#" aria-label="Close" .onclick=${this.clickCloseSayt}>${this.closeText}</a>
     `;
   }
 }
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -19,7 +19,7 @@ export default class Sayt extends LitElement {
   /**
    * Stores the ID of the relevant search element.
    */
-  @property({ type: String, reflect: true }) searchbar = '';
+  @property({ type: String, reflect: true }) searchbox = '';
   /**
    * Customizes the text in the close button.
    */
@@ -97,7 +97,7 @@ export default class Sayt extends LitElement {
   /**
    * Makes SAYT visible if the event refers to the correct SAYT component.
    *
-   * @param event An event that can contain a searchbar ID.
+   * @param event An event that can contain a searchbox ID.
    */
   showCorrectSayt(event: CustomEvent) {
     if (this.isCorrectSayt(event)) {
@@ -115,7 +115,7 @@ export default class Sayt extends LitElement {
   /**
    * Hides SAYT if the event refers to the correct SAYT component.
    *
-   * @param event An event that can contain a searchbar ID.
+   * @param event An event that can contain a searchbox ID.
    */
   hideCorrectSayt(event: CustomEvent) {
     if (this.isCorrectSayt(event)) {
@@ -125,16 +125,16 @@ export default class Sayt extends LitElement {
 
   /**
    * Determines whether an event refers to the correct SAYT. This is true if
-   * a matching `searchbar` ID is specified, if the event has no `searchbar`
-   * ID specified, or if SAYT has no `searchbar` ID specified.
+   * a matching `searchbox` ID is specified, if the event has no `searchbox`
+   * ID specified, or if SAYT has no `searchbox` ID specified.
    *
-   * @param event An event that can contain a searchbar ID for comparison.
+   * @param event An event that can contain a searchbox ID for comparison.
    */
   isCorrectSayt(event: CustomEvent) {
-    const searchbar = event.detail && event.detail.searchbar;
-    return this.searchbar === undefined ||
-      searchbar === this.searchbar ||
-      searchbar === undefined;
+    const searchbox = event.detail && event.detail.searchbox;
+    return this.searchbox === undefined ||
+      searchbox === this.searchbox ||
+      searchbox === undefined;
   }
 
   /**
@@ -165,8 +165,8 @@ export default class Sayt extends LitElement {
    * @param node The node to check for containment.
    */
   nodeInSearchBar(node: Node) {
-    if (!this.searchbar) return false;
-    const searchBar = document.querySelector('#' + this.searchbar);
+    if (!this.searchbox) return false;
+    const searchBar = document.querySelector('#' + this.searchbox);
     return !!searchBar && searchBar.contains(node);
   }
 
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -38,7 +38,7 @@ export default class Sayt extends LitElement {
     this.showSayt = this.showSayt.bind(this);
     this.hideSayt = this.hideSayt.bind(this);
     this.processClick = this.processClick.bind(this);
-    this.processKeyPress = this.processKeyPress.bind(this);
+    this.processKeyEvent = this.processKeyEvent.bind(this);
     this.nodeInSearchbox = this.nodeInSearchbox.bind(this);
     this.hideCorrectSayt = this.hideCorrectSayt.bind(this);
     this.showCorrectSayt = this.showCorrectSayt.bind(this);
@@ -55,7 +55,7 @@ export default class Sayt extends LitElement {
     window.addEventListener(AUTOCOMPLETE_RECEIVED_RESULTS_EVENT, this.showCorrectSayt);
     window.addEventListener(SAYT_EVENT.SAYT_HIDE, this.hideCorrectSayt);
     window.addEventListener('click', this.processClick);
-    window.addEventListener('keydown', this.processKeyPress);
+    window.addEventListener('keydown', this.processKeyEvent);
   }
 
   /**
@@ -68,7 +68,7 @@ export default class Sayt extends LitElement {
     window.removeEventListener(AUTOCOMPLETE_RECEIVED_RESULTS_EVENT, this.showCorrectSayt);
     window.removeEventListener(SAYT_EVENT.SAYT_HIDE, this.hideCorrectSayt);
     window.removeEventListener('click', this.processClick);
-    window.removeEventListener('keydown', this.processKeyPress);
+    window.removeEventListener('keydown', this.processKeyEvent);
   }
 
   createRenderRoot() {
@@ -175,7 +175,7 @@ export default class Sayt extends LitElement {
    *
    * @param event A keyboard event used for checking which key has been pressed.
    */
-  processKeyPress(event: KeyboardEvent) {
+  processKeyEvent(event: KeyboardEvent) {
     if (event.key === 'Escape') {
       this.hideSayt();
     }
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -186,8 +186,8 @@ export default class Sayt extends LitElement {
   render() {
     return html`
       ${ this.showCloseButton ?
-        html`<a href="#" aria-label="Close" .onclick=${this.clickCloseSayt}>
-          ${this.closeText}
+        html`<a href="#" aria-label="Close" .onclick=${ this.clickCloseSayt }>
+          ${ this.closeText }
         </a>`
         : ``
       }
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -162,6 +162,7 @@ export default class Sayt extends LitElement {
    * @param node The node to check for containment.
    */
   nodeInSearchBar(node: Node) {
+    if (!this.searchbar) return false;
     const searchBar = document.querySelector('#' + this.searchbar);
     return !!searchBar && searchBar.contains(node);
   }
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -23,11 +23,11 @@ export default class Sayt extends LitElement {
   /**
    * Customizes the text in the close button.
    */
-  @property({ type: String }) closeText = 'Close';
+  @property({ type: String, reflect: true }) closeText = 'Close';
   /**
    * Shows a button to allow for closing SAYT manually.
    */
-  @property({ type: Boolean }) showCloseButton = false;
+  @property({ type: Boolean, reflect: true }) showCloseButton = false;
 
   /**
    * Calls superclass constructor and bind methods.
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -133,9 +133,7 @@ export default class Sayt extends LitElement {
    */
   isCorrectSayt(event: CustomEvent): boolean {
     const searchbox = event.detail && event.detail.searchbox;
-    return this.searchbox === undefined ||
-      searchbox === this.searchbox ||
-      searchbox === undefined;
+    return !searchbox || !this.searchbox || searchbox === this.searchbox;
   }
 
   /**
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -55,7 +55,7 @@ export default class Sayt extends LitElement {
     window.addEventListener(AUTOCOMPLETE_RECEIVED_RESULTS_EVENT, this.showCorrectSayt);
     window.addEventListener(SAYT_EVENT.SAYT_HIDE, this.hideCorrectSayt);
     window.addEventListener('click', this.processClick);
-    window.addEventListener('keypress', this.processKeyPress);
+    window.addEventListener('keydown', this.processKeyPress);
   }
 
   /**
@@ -68,7 +68,7 @@ export default class Sayt extends LitElement {
     window.removeEventListener(AUTOCOMPLETE_RECEIVED_RESULTS_EVENT, this.showCorrectSayt);
     window.removeEventListener(SAYT_EVENT.SAYT_HIDE, this.hideCorrectSayt);
     window.removeEventListener('click', this.processClick);
-    window.removeEventListener('keypress', this.processKeyPress);
+    window.removeEventListener('keydown', this.processKeyPress);
   }
 
   createRenderRoot() {
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -113,14 +113,16 @@ export default class Sayt extends LitElement {
 
   /**
    * Determines whether an event refers to the correct SAYT. This is true if
-   * either a matching `searchbar` ID is specified, or if no `searchbar` ID
-   * is specified at all.
+   * a matching `searchbar` ID is specified, if the event has no `searchbar`
+   * ID specified, or if SAYT has no `searchbar` ID specified.
    *
    * @param event An event that can contain a searchbar ID for comparison.
    */
   isCorrectSayt(event: CustomEvent) {
     const searchbar = event.detail.searchbar;
-    return searchbar === this.searchbar || searchbar === undefined;
+    return this.searchbar === undefined ||
+      searchbar === this.searchbar ||
+      searchbar === undefined;
   }
 
   /**
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -39,7 +39,7 @@ export default class Sayt extends LitElement {
     this.hideSayt = this.hideSayt.bind(this);
     this.processClick = this.processClick.bind(this);
     this.processKeyPress = this.processKeyPress.bind(this);
-    this.nodeInSearchBar = this.nodeInSearchBar.bind(this);
+    this.nodeInSearchbox = this.nodeInSearchbox.bind(this);
     this.hideCorrectSayt = this.hideCorrectSayt.bind(this);
     this.showCorrectSayt = this.showCorrectSayt.bind(this);
     this.isCorrectSayt = this.isCorrectSayt.bind(this);
@@ -140,9 +140,9 @@ export default class Sayt extends LitElement {
    * 
    * @param event The click event.
    */
-  processClick(event: Event) {
+  processClick(event: MouseEvent) {
     const target = event.target as Node;
-    if (this.contains(target) || this.nodeInSearchBar(target)) return;
+    if (this.contains(target) || this.nodeInSearchbox(target)) return;
 
     this.hideSayt();
   }
@@ -162,15 +162,18 @@ export default class Sayt extends LitElement {
    * 
    * @param node The node to check for containment.
    */
-  nodeInSearchBar(node: Node): boolean {
+  nodeInSearchbox(node: Node): boolean {
     if (!this.searchbox) return false;
-    const searchBar = document.getElementById(this.searchbox);
-    return !!searchBar && searchBar.contains(node);
+    const searchbox = document.getElementById(this.searchbox);
+    return !!searchbox && searchbox.contains(node);
   }
 
   /**
-   * Processes a keypress event in order to close SAYT under the right conditions.
-   * @param event 
+   * Processes a keyboard event in order to close SAYT when certain keys are pressed.
+   * Namely:
+   *   - Escape
+   *
+   * @param event A keyboard event used for checking which key has been pressed.
    */
   processKeyPress(event: KeyboardEvent) {
     if (event.key === 'Escape') {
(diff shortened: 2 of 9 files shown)


.storybook/common.ts
```diff
@@ -0,0 +1,16 @@
...
+export function getDisplayCode(code: string): string {
+    return `
+      <style>
+        pre.code-output {
+          padding: 15px;
+          background-color: #EEEEEE;
+        }
+      </style>
+      <h3>The code</h3>
+      <pre class="code-output"><code>${entities.encode(code)}</code></pre>
```
ssbelle: spaces inside {} here

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -64,15 +94,103 @@ export default class Sayt extends LitElement {
...
+    }
+  }
+
+  /**
+   * Returns a TemplateResult object for rendering in LitElement.
+   */
   render() {
     return html`
+      ${ this.showCloseButton ?
+        html`<a href="#" aria-label="Close" .onclick=${this.clickCloseSayt}>
```
ssbelle: I don't think this . should be before the onclick?

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -64,15 +94,103 @@ export default class Sayt extends LitElement {
...
+          ${this.closeText}
```
ssbelle: spaces inside {} ?

packages/web-components/@sfx/sayt/stories/index.ts
```diff
@@ -1,25 +1,48 @@
 import { storiesOf } from '@storybook/html';
 import { withKnobs, text, boolean } from '@storybook/addon-knobs';
+
```
ssbelle: remove extra space ?

packages/web-components/@sfx/sayt/stories/index.ts
```diff
@@ -1,25 +1,48 @@
...
+// story's events to not affect another story.
+const autocompleteDataReceivedEvent = new CustomEvent('sfx::autocomplete_received_results',
   { detail: [
       { "title": "Brands", "items": [{ "label": "Cats" }, { "label": "Dogs" }] },
       { "title": "default", "items": [{ "label": "Cars" }, { "label": "Bikes" }] }
     ],
     bubbles: true }
 );
 
+function getSayt(searchbar='', showSayt=true): string {
```
ssbelle: should there be spacing after the = here?

packages/web-components/@sfx/sayt/stories/index.ts
```diff
@@ -1,25 +1,48 @@
...
+  const showAttribute = boolean('visible', showSayt) ? 'visible' : '';
+  const closeText = text('Close link text', 'Close');
+  const showCloseButton = boolean('Show Close button', true) ? 'showclosebutton' : '';
+
+  return `<sfx-sayt${searchbar ? ` searchbar="${searchbar}"` : ''}
```
ssbelle: spaces after { and before }

packages/web-components/@sfx/sayt/stories/index.ts
```diff
@@ -1,25 +1,48 @@
...
+  closetext="${closeText}"${showCloseButton ? `
```
ssbelle: spaces for {}

packages/web-components/@sfx/sayt/stories/index.ts
```diff
@@ -1,25 +1,48 @@
...
   .add('Default', () => {
-    const showAttribute = boolean('visible', false) ? 'visible' : '';
+    emitEventInFuture(autocompleteDataReceivedEvent, 100);
 
-    setTimeout(() => {
-      window.dispatchEvent(autocompleteDataReceivedEvent);
-    }, 100);
+    const sayt = getSayt();
+    return `
+      ${sayt}
```
ssbelle: spaces?

packages/web-components/@sfx/sayt/stories/index.ts
```diff
@@ -1,25 +1,48 @@
...
 
-    return `<sfx-sayt ${ showAttribute }></sfx-sayt>`
+      ${getDisplayCode(sayt)}
```
ssbelle: same here

packages/web-components/@sfx/sayt/stories/index.ts
```diff
@@ -30,28 +53,72 @@ storiesOf('Components|SAYT', module)
...
+  .add('Responding to Events - sayt_hide & sayt_show ', () => {
+    emitEventInFuture(autocompleteDataReceivedEvent, 100);
+    emitEventInFuture(new Event('sfx::sayt_hide'), 2000);
+    emitEventInFuture(new Event('sfx::sayt_show'), 4000);
 
+    const sayt = getSayt('', false);
     return `
-      <sfx-sayt></sfx-sayt>
+      ${sayt}
+      ${getDisplayCode(sayt)}
```
ssbelle: spaces here and above?

packages/web-components/@sfx/sayt/stories/index.ts
```diff
@@ -30,28 +53,72 @@ storiesOf('Components|SAYT', module)
...
       `
     }
   })
+  .add('SAYT with simple search input', () => {
+    emitEventInFuture(autocompleteDataReceivedEvent, 100);
+
+    const input = `<input type="text" id="search-bar" placeholder="Search here" />`;
+    const sayt = getSayt();
+    return `
+      ${input}
```
ssbelle: indentation, spacing with {} and semi colons from here to line 122?

packages/web-components/@sfx/sayt/stories/index.ts
ssbelle: also maybespace after the tag opening and or bring the first ${searchbar onto the next line? looks cramped

packages/web-components/@sfx/sayt/stories/index.ts
DanielMoniz-GBI: I left this here on purpose to separate 3rd party imports from local ones, but I can do whatever the team prefers.

packages/web-components/@sfx/sayt/stories/index.ts
DanielMoniz-GBI: In my opinion: no, but that doesn't mean it's cool to do in JavaScript. Will let someone else add their input.

packages/web-components/@sfx/sayt/stories/index.ts
DanielMoniz-GBI: This looks very cramped because I need to keep whitespace to a minimum. It will be going into a `<pre>` tag, where whitespace is respected, so any extra newlines inside the string cause newlines to appear on the page.

packages/web-components/@sfx/sayt/stories/index.ts
DanielMoniz-GBI: I have to be careful with indentation because it can cause the displayed lines of code to be mis-indented.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -28,8 +52,11 @@ export default class Sayt extends LitElement {
   connectedCallback() {
     super.connectedCallback();
 
-    window.addEventListener(SAYT_EVENT.SAYT_SHOW, this.showSayt);
-    window.addEventListener(SAYT_EVENT.SAYT_HIDE, this.hideSayt);
+    window.addEventListener(SAYT_EVENT.SAYT_SHOW, this.showCorrectSayt);
+    window.addEventListener(AUTOCOMPLETE_RECEIVED_RESULTS_EVENT, this.showCorrectSayt);
+    window.addEventListener(SAYT_EVENT.SAYT_HIDE, this.hideCorrectSayt);
+    window.addEventListener('click', this.processClick);
+    window.addEventListener('keypress', this.processKeyPress);
```
jmbtutor: I think it's more usual to act on `keydown` for situations like these.

What's your opinion on having these listeners around only when sayt is visible?

packages/web-components/@sfx/sayt/stories/index.ts
```diff
@@ -1,25 +1,48 @@
 import { storiesOf } from '@storybook/html';
 import { withKnobs, text, boolean } from '@storybook/addon-knobs';
+
 import '../src/index.ts';
+import { getDisplayCode } from '../../../../../.storybook/common';
```
jmbtutor: Do we not have a package-local util file that re-exports entities from `.storybook` so we don't have to go all the way up for them? If we can't have multiple story files, then ignore this suggestion.

packages/web-components/@sfx/sayt/stories/index.ts
jmbtutor: I'm in favour of not having spaces because you can clearly see which ones are local and which ones are not based on the path, but I haven't been flagging them since we haven't had the discussion yet.

packages/web-components/@sfx/sayt/stories/index.ts
```diff
@@ -1,25 +1,48 @@
 import { storiesOf } from '@storybook/html';
 import { withKnobs, text, boolean } from '@storybook/addon-knobs';
+
 import '../src/index.ts';
```
jmbtutor: You didn't change this, but please remove the extension.

packages/web-components/@sfx/sayt/stories/index.ts
jmbtutor: I expect that our linter settings will require padding.

packages/web-components/@sfx/sayt/stories/index.ts
```diff
@@ -1,25 +1,48 @@
...
 
+function getSayt(searchbar='', showSayt=true): string {
+  const showAttribute = boolean('visible', showSayt) ? 'visible' : '';
+  const closeText = text('Close link text', 'Close');
+  const showCloseButton = boolean('Show Close button', true) ? 'showclosebutton' : '';
+
+  return `<sfx-sayt${ searchbar ? ` searchbar="${ searchbar }"` : '' }
+  closetext="${ closeText }"${ showCloseButton ? `
+  ${ showCloseButton }` : ''}${ showAttribute ? `
+  ${ showAttribute }` : ''}
```
jmbtutor: Is there a way we can format this better?

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -16,8 +17,11 @@ describe('Sayt Component', () => {
 
       sayt.connectedCallback();
 
-      expect(addEventListener).to.be.calledWith(SAYT_EVENT.SAYT_SHOW, sayt.showSayt);
-      expect(addEventListener).to.be.calledWith(SAYT_EVENT.SAYT_HIDE, sayt.hideSayt);
+      expect(addEventListener).to.be.calledWith(SAYT_EVENT.SAYT_SHOW, sayt.showCorrectSayt);
+      expect(addEventListener).to.be.calledWith(SAYT_EVENT.SAYT_HIDE, sayt.hideCorrectSayt);
+      expect(addEventListener).to.be.calledWith('click', sayt.processClick);
+      expect(addEventListener).to.be.calledWith('keypress', sayt.processKeyPress);
+      expect(addEventListener).to.be.calledWith(AUTOCOMPLETE_RECEIVED_RESULTS_EVENT);
```
jmbtutor: This should assert the callback that's passed as well.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -64,15 +94,103 @@ export default class Sayt extends LitElement {
...
+  }
+
+  /**
+   * Determines whether an event refers to the correct SAYT. This is true if
+   * a matching `searchbar` ID is specified, if the event has no `searchbar`
+   * ID specified, or if SAYT has no `searchbar` ID specified.
+   *
+   * @param event An event that can contain a searchbar ID for comparison.
+   */
+  isCorrectSayt(event: CustomEvent) {
```
ChantalDesrochers: Should the return value type be indicated ('boolean'), or is it unnecessary?

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+
   describe('render()', () => {
     it('should return an instance of TemplateResult', () => {
       const result = sayt.render();
 
       expect(result).to.be.an.instanceof(TemplateResult);
     });
   });
+
+  describe('processClick()', () => {
```
jmbtutor: I would order these tests in the order they appear in the file.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -64,15 +94,103 @@ export default class Sayt extends LitElement {
...
+   */
+  processClick(event: Event) {
+    const target = event.target as Node;
+    if (this.contains(target) || this.nodeInSearchBar(target)) return;
+
+    this.hideSayt();
+  }
+
+  /**
+   * Handles hiding SAYT on click on a close link/button (or other event).
```
ChantalDesrochers: Should this say 'Handles hiding SAYT on click _of_ a close link/button' ?

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+    let node: any = 'some-node';
```
jmbtutor: If `target` is a `Node`, then `node` at minimum should be an object value so that the basic type is the same. If this is an object, then you'll have to set it to a new object every time in `beforeEach`.

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+    let event: Event;
```
jmbtutor: We can type this as `any` to avoid casts.

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+
+    beforeEach(() => {
+      event = {
+        target: node,
+      } as Event;
+      sayt.nodeInSearchBar = () => false;
```
jmbtutor: I think we've been using stubs for these.
```suggestion
      stub(sayt, 'nodeInSearchBar').returns(false);
```

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+    });
+
+    it('should hide SAYT if the event target is nowhere relevant', () => {
+      sayt.contains = stub().returns(false);
```
jmbtutor:
```suggestion
      stub(sayt, 'contains').returns(false);
```

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -64,15 +94,103 @@ export default class Sayt extends LitElement {
...
+   *
+   * @param event An event with a default action to be prevented.
+   */
+  clickCloseSayt(event: Event) {
+    event.preventDefault();
+    this.hideSayt();
```
ChantalDesrochers: Currently this function doesn't check for the correct sayt, if there are multiple on a page. Should some check be incorporated?

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -64,15 +94,103 @@ export default class Sayt extends LitElement {
...
+  }
+
+  /**
+   * Checks whether a given node is inside of SAYT's identified search bar.
+   * 
+   * @param node The node to check for containment.
+   */
+  nodeInSearchBar(node: Node) {
```
ChantalDesrochers: Should we include return type? (boolean)

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+      sayt.hideSayt = spy();
```
jmbtutor: We don't want it to actually do the action, so this should be a stub.

This and the above comments apply to the rest of the occurrences.

```suggestion
      stub(sayt, 'hideSayt');
```

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+      expect(sayt.hideSayt).to.not.be.called;
+    });
+  });
+
+  describe('nodeInSearchBar()', () => {
+    it('should return true if given node is contained in the search bar', () => {
+      const searchbar = {
+        contains: stub().returns(true),
+      };
+      const querySelector = stub(document, 'querySelector').callsFake(() => searchbar);
```
jmbtutor: This line will probably change with the corresponding suggestion in the code under test.

```suggestion
      const querySelector = stub(document, 'querySelector').returns(searchbar);
```

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+      expect(sayt.contains).to.be.calledWith(node);
+      expect(sayt.nodeInSearchBar).to.be.calledWith(node);
+      expect(sayt.hideSayt).to.not.be.called;
+    });
+  });
+
+  describe('nodeInSearchBar()', () => {
+    it('should return true if given node is contained in the search bar', () => {
+      const searchbar = {
+        contains: stub().returns(true),
```
jmbtutor: What's your opinion on `stub().returns(true)` vs `spy(() => true)`?

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+      };
+      const querySelector = stub(document, 'querySelector').callsFake(() => searchbar);
+      sayt.searchbar = 'searchbar-id';
+
+      const result = sayt.nodeInSearchBar('node');
+
+      expect(querySelector).to.be.calledWith('#searchbar-id');
+      expect(searchbar.contains).to.be.calledWith('node');
+      expect(result).to.equal(true);
```
jmbtutor:
```suggestion
      expect(result).to.be.true;
```

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+    });
+
+    it('should return false if given node is not contained in the search bar', () => {
```
jmbtutor: Similar comments as in the test above.

packages/web-components/@sfx/sayt/stories/index.ts
ChantalDesrochers: Not sure if it matters, but should we keep the import types grouped together/or in a certain order? (single imports vs. files)

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+
+    it('should return false if there is no search bar on the page', () => {
+      const querySelector = stub(document, 'querySelector').returns(null);
+
+      const result = sayt.nodeInSearchBar('node');
+
+      expect(result).to.be.false;
+    });
+
+    it('should not call document.querySelector() if this.searchbar is not set', () => {
```
jmbtutor: I think this is an implementation detail. Your test should probably be that it returns `false` if ~~searchbar~~ searchbox is not set.

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+      const querySelector = stub(document, 'querySelector');
+
+      sayt.nodeInSearchBar('node');
+
+      expect(querySelector).to.not.be.called;
+    });
+  });
+
+  describe('processKeyPress()', () => {
+    it('should exist as a function', () => {
```
jmbtutor: This test is not necessary.

packages/web-components/@sfx/sayt/stories/index.ts
```diff
@@ -1,25 +1,48 @@
...
 import '../src/index.ts';
+import { getDisplayCode } from '../../../../../.storybook/common';
 
-const autocompleteDataReceivedEvent = new CustomEvent('autocomplete_received_results',
+// @TODO allow for sending event with searchbar ID. This should allow for one
+// story's events to not affect another story.
+const autocompleteDataReceivedEvent = new CustomEvent('sfx::autocomplete_received_results',
   { detail: [
       { "title": "Brands", "items": [{ "label": "Cats" }, { "label": "Dogs" }] },
       { "title": "default", "items": [{ "label": "Cars" }, { "label": "Bikes" }] }
```
ChantalDesrochers: The autocomplete component functions on the title property being empty rather than the string 'default'
{ "title": "", "items": [{ "label": "Cars" }, { "label": "Bikes" }] }
instead of 
{ "title": "default", "items": [{ "label": "Cars" }, { "label": "Bikes" }] }

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+      expect(sayt.processKeyPress).to.be.a('function');
+    });
+
+    it('should hide SAYT when pressing escape', () => {
+      const event = { key: "Escape" } as KeyboardEvent;
```
jmbtutor: You can type this as `any` to avoid casting:

```typescript
      const event: any = { key: "Escape" };
```

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+
+      sayt.processKeyPress(event);
+
+      expect(sayt.hideSayt).to.be.called;
+    });
+
+    it('should not hide SAYT when pressing any character other than escape', () => {
+      const event = { key: "j" } as KeyboardEvent;
+      const event2 = { key: "Enter" } as KeyboardEvent;
+      const event3 = { key: "Space" } as KeyboardEvent;
```
jmbtutor: You can type these as `any` to avoid casts.

packages/web-components/@sfx/sayt/stories/index.ts
ChantalDesrochers: I've only worked on one project to reference, but we **did** include spacing.
I also checked MDN, and the example they provided includes spacing.

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+  });
+
+  describe('processKeyPress()', () => {
+    it('should exist as a function', () => {
+      expect(sayt.processKeyPress).to.be.a('function');
+    });
+
+    it('should hide SAYT when pressing escape', () => {
+      const event = { key: "Escape" } as KeyboardEvent;
+      sayt.hideSayt = spy();
```
jmbtutor: Same thing about this being a stub. I won't flag the rest of the occurrences, but it applies to all of them as well.

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+      sayt.processKeyPress(event);
+      sayt.processKeyPress(event2);
+      sayt.processKeyPress(event3);
+
+      expect(sayt.hideSayt).to.not.be.called;
+    });
+  });
+
+  describe('clickCloseSayt()', () => {
+    it('should close SAYT when activated', () => {
```
jmbtutor: "when activated" in the sense of "when the function is called" is implied. Having it in the description suggests to me that it's possible for SAYT not to be "activated". 

```suggestion
    it('should close SAYT', () => {
```

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+      const hideSayt = stub(sayt, 'hideSayt');
+      const event = { preventDefault: () => null };
+
+      sayt.clickCloseSayt(event);
+
+      expect(hideSayt).to.be.calledOnce;
+    });
+
+    it('should prevent the default event action when activated', () => {
+      stub(sayt, 'hideSayt');
```
jmbtutor: I would move this below the `const`s.

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,11 +72,244 @@ describe('Sayt Component', () => {
...
+    it('should close SAYT when activated', () => {
+      const hideSayt = stub(sayt, 'hideSayt');
+      const event = { preventDefault: () => null };
+
+      sayt.clickCloseSayt(event);
+
+      expect(hideSayt).to.be.calledOnce;
+    });
+
+    it('should prevent the default event action when activated', () => {
```
jmbtutor: Same comment as above.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -64,15 +94,103 @@ export default class Sayt extends LitElement {
...
+    }
+  }
+
+  /**
+   * Returns a TemplateResult object for rendering in LitElement.
+   */
   render() {
     return html`
+      ${ this.showCloseButton ?
+        html`<a href="#" aria-label="Close" .onclick=${ this.clickCloseSayt }>
```
ChantalDesrochers: Lit-Element has @event bindings. We're currently using @click, @input, & @keydown in the autocomplete and search-box components.

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -133,10 +132,26 @@ export default class SearchBox extends Base {
...
     this.dispatchEvent(searchBoxClickedEvent);
   }
 
+  /**
+   * Returns an event with a standard structure.
+   *
+   * @param type The type (or name) of the event to be emitted.
+   * @param detail A payload to be sent with the event.
+   */
+  getCustomEvent(type: string, detail: object = {}) {
```
jmbtutor: I would call this `createCustomEvent`.

jmbtutor: I think `detail` should be optional. Attempting to spread `undefined` will succeed and expand to nothing.

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -133,10 +132,26 @@ export default class SearchBox extends Base {
...
+    return new CustomEvent(type, {
+      detail: {
+        ...detail,
+        searchbox: this.id,
+      },
+      bubbles: true,
+    })
```
jmbtutor:
```suggestion
    });
```

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
```diff
@@ -68,36 +68,37 @@ describe('SearchBox Component', () => {
...
 
-      expect(searchboxDispatchEvent).to.have.been.calledWith(searchboxClearedEvent);
+      expect(getCustomEvent).to.be.calledWith(SEARCHBOX_EVENT.SEARCHBOX_CLEAR_CLICK);
+      expect(searchboxDispatchEvent).to.be.calledWith('some-event');
     });
   });
 
   describe('updateText', () => {
-    it('should update the value property in response to the data received', () => {
+    it('should update the value property with information from the event', () => {
```
jmbtutor:
```suggestion
    it('should update the value property with data from the event', () => {
```

packages/web-components/@sfx/sayt/stories/index.ts
jmbtutor: For interpolated strings, we generally haven't been putting padding inside the interpolated portion.

packages/web-components/@sfx/sayt/src/sayt.ts
DanielMoniz-GBI: Two of them make SAYT visible, so those have to exist when SAYT is not visible. They technically don't need to exists when it is.

The others can presumably exist only in the opposite case. I'm unaware of what impact this would have on performance. I'm tending toward leaving this for now and optimizing later. Let's discuss this in person.

DanielMoniz-GBI: Additionally, why use `keypress`? This at least guarantees that any `keydown` events will have fired. Is it a UX thing?

packages/web-components/@sfx/sayt/stories/index.ts
jmbtutor: The order I've been suggesting is non-local imports (in some order that makes sense; alphabetical if you can't think of one) followed by local imports. For tests: non-local imports, test utilities, the class under test, and then other supporting imports. All imports should be in one block (i.e. no spaces between imports).

packages/web-components/@sfx/sayt/src/sayt.ts
DanielMoniz-GBI: It should be there! Good catch.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -64,15 +94,103 @@ export default class Sayt extends LitElement {
...
+      searchbar === undefined;
+  }
+
+  /**
+   * Processes a click event in order to close SAYT under the right conditions.
+   * 
+   * @param event The click event.
+   */
+  processClick(event: Event) {
+    const target = event.target as Node;
```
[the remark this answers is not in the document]
DanielMoniz-GBI: I think this is here because I don't seem to have access to a `ClickEvent`, despite the fact that I have access to a `KeyboardEvent`. I get compile errors.

packages/web-components/@sfx/sayt/src/sayt.ts
DanielMoniz-GBI: Each SAYT component registers a listener for clicking its own Close button. We don't need any additional check for that reason.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -64,15 +94,103 @@ export default class Sayt extends LitElement {
...
+  }
+
+  /**
+   * Checks whether a given node is inside of SAYT's identified search bar.
+   * 
+   * @param node The node to check for containment.
+   */
+  nodeInSearchBar(node: Node) {
+    if (!this.searchbar) return false;
+    const searchBar = document.querySelector('#' + this.searchbar);
```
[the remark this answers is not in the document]
DanielMoniz-GBI: Sometimes I surprise even myself!

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -64,15 +94,103 @@ export default class Sayt extends LitElement {
...
+    return !!searchBar && searchBar.contains(node);
```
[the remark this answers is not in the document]
DanielMoniz-GBI: I believe it does.
```
document.contains(document);
=> true
```

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -64,15 +94,103 @@ export default class Sayt extends LitElement {
...
+  }
+
+  /**
+   * Processes a keypress event in order to close SAYT under the right conditions.
+   * @param event 
+   */
+  processKeyPress(event: KeyboardEvent) {
+    if (event.key === "Escape") {
```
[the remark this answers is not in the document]
DanielMoniz-GBI: Any reason for this change?

packages/web-components/@sfx/sayt/src/sayt.ts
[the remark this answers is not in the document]
DanielMoniz-GBI: Honestly, I'm not sure. They can presumably style a link as a button if they choose. I guess there is no real fallback from there being no JavaScript, though, so the link isn't providing any value over a button.

packages/web-components/@sfx/sayt/src/sayt.ts
JoeyDeol: `keypress` has been deprecated according to mdn: https://developer.mozilla.org/en-US/docs/Web/API/Document/keypress_event. 

They suggest using `keydown` or `beforeinput` instead.

packages/web-components/@sfx/sayt/stories/index.ts
DanielMoniz-GBI: The import order is normally based on imports being 1) built-in, 2) third-party, and 3) project files. I am switching the order slightly for prettiness, though.

packages/web-components/@sfx/sayt/stories/index.ts
DanielMoniz-GBI: I don't think so, but I'm open to suggestions. The way it's done here is very intentional.

It's possible to use HTML comments to allow for code line breaks without visual line breaks, but I don't really want to inject comments all over the code (especially since it will be output on each story).

jmbtutor: Using escape sequences (e.g. `\n`) might help.

DanielMoniz-GBI: The problem is too much whitespace, rather than not enough. I can use `\n` to produce empty lines between each piece, but I'll end up putting most of it on one line.

As soon as one attribute ends, I need to begin a new one to prevent extra newlines being added.

If you have an idea, feel free to take a shot at it, but I won't spend additional time on it unless I have a concrete plan.

packages/web-components/@sfx/sayt/test/sayt.test.ts
DanielMoniz-GBI: Why does this matter? According to the tests, it doesn't matter what node is. We are just asserting that it is passed along.

packages/web-components/@sfx/sayt/test/sayt.test.ts
DanielMoniz-GBI: I will make this change because I think an assertion could be made that `.nodeInSearchBar()` is actually called (which I will add). In the current state of the code, though, there is no usage of that functionality. This was an intentional choice.

packages/web-components/@sfx/sayt/test/sayt.test.ts
DanielMoniz-GBI: To be clear, your suggestion is not one of function, but one of convention. This also prevents the action from taking place. I will make the change(s).

packages/web-components/@sfx/sayt/test/sayt.test.ts
DanielMoniz-GBI: Your argument above is to use `stub()` instead of `spy()`, though the difference seems to be that we are building our own object here rather than modifying an existing object (of whose function calls we are trying to restrict).

I would say that, more generally, we should probably be using spies if we are building out our own mock objects and we additionally need to track their usage. I'll switch to `spy(() => true)` here (which I did not know about), and I'll use that convention going forward. I don't want to be too pedantic about its use on others, though - the distinction is pretty specific.

packages/web-components/@sfx/search-box/test/unit/search-box.test.ts
DanielMoniz-GBI: Yup.

packages/web-components/@sfx/sayt/src/sayt.ts
DanielMoniz-GBI: Thanks, @JoeyDeol, I didn't realize it was deprecated.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -64,15 +94,103 @@ export default class Sayt extends LitElement {
...
   hideSayt() {
     this.visible = false;
   }
 
+  /**
+   * Hides SAYT if the event refers to the correct SAYT component.
+   *
+   * @param event An event that can contain a searchbar ID.
+   */
+  hideCorrectSayt(event: CustomEvent) {
```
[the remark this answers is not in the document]
DanielMoniz-GBI: I'm not sure what you mean. It doesn't currently attempt to return anything - its only job is the side effect of potentially hiding SAYT.

packages/web-components/@sfx/sayt/stories/index.ts
jmbtutor: Option 1: "hide the whitespace in the interpolated area"

```js
  return \
`<sfx-sayt${
  searchbox ? `\n  searchbox="${searchbox}"` : ''}
  closetext="${closeText}"${
  showCloseButton ? `\n  ${showCloseButton}` : ''}${
  showAttribute ? `\n  ${showAttribute}` : ''}
></sfx-sayt>`;
```

Option 2: "string concatenation is your friend"

```js
return \
'<sfx-sayt\n'
+ (searchbox ? `  searchbox="${searchbox}"\n` : '')
+ `  closetext="${closeText}"\n`
+ (showCloseButton ? `  ${showClosebutton}\n` : '')
+ (showAttribute ? `  ${showAttribute}\n` : '')
+ '></sfx-sayt>'
```

I'm somewhat partial to option 2, although it doesn't show the indentation as well as option 1.

packages/web-components/@sfx/sayt/stories/index.ts
```diff
@@ -30,28 +52,86 @@ storiesOf('Components|SAYT', module)
...
+        - Show automatically once sub-component Autocomplete receives results.
+        - Receiving sayt_hide event after 2 seconds.
+        - Receiving sayt_show event after 4 seconds.
+      `
+    }
+  })
+  .add('SAYT with simple search input', () => {
+    emitEventInFuture(autocompleteDataReceivedEvent, 100);
+
+    const input = `<input type="text" id="search-bar" placeholder="Search here" />`;
```
jmbtutor: Remember to change anything that's relying on this ID as well.

```suggestion
    const input = `<input type="text" id="search-box" placeholder="Search here" />`;
```

packages/web-components/@sfx/sayt/stories/index.ts
```diff
@@ -30,28 +52,86 @@ storiesOf('Components|SAYT', module)
...
+    const sayt = getSayt();
+    return `
+      ${ input }
+      <br />
+      ${ sayt }
+      ${ getDisplayCode(`${ input }
+${ sayt }`) }
```
jmbtutor: Personal opinion, but this is a situation where I would prefer an escape sequence since the embedded newline hinders readability:

```js
getDisplayCode(`${ input }\n${ sayt }`)
```

packages/web-components/@sfx/sayt/stories/index.ts
```diff
@@ -30,28 +52,86 @@ storiesOf('Components|SAYT', module)
...
       `
     }
   })
+  .add('SAYT with multiple search inputs', () => {
+    emitEventInFuture(autocompleteDataReceivedEvent, 100);
+
+    const input1 = `<input type="text" id="search-bar1" placeholder="Search here" />`;
+    const input2 = `<input type="text" id="search-bar2" placeholder="Or search here" />`;
+    const sayt1 = getSayt('search-bar1');
+    const sayt2 = getSayt('search-bar2');
```
jmbtutor: "search-box"

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,6 +96,210 @@ describe('Sayt Component', () => {
...
+      const event = { detail: { searchbox } };
+
+      const result = sayt.isCorrectSayt(event);
+
+      expect(result).to.be.true;
+    });
+
+    it('should return true if event does not provide a searchbox ID', () => {
+      sayt.searchbox = 'some-searchbox-id';
+      const event = { detail: {} };
```
jmbtutor: Unless there's a compelling reason otherwise, I would put declarations and initializers above other assignments. At least, that's what we've been doing in the other tests.

packages/web-components/@sfx/sayt/test/sayt.test.ts
jmbtutor: There's a difference between objects and primitives in that the former is passed by reference, whereas the latter is passed by value. If you have a primitive where an object is expected, then you lose the ability to convincingly test whether it's the exact value or if it just has similar contents because both is the same for a primitive. In other words, you lose information by using a primitive. Doing this _may_ cause us to miss a bug that would otherwise have been caught by using an object.

Does `node` need to be a full `Node` object? Of course not, but I think it should still be an object, even if it's empty.

Furthermore, to paraphrase Uncle Bob, unit tests can be seen as a type of code sample, one that shows all the different ways that the class can be called. The samples are so correct and in-sync with the "production" code that you can run them. Having a string here signals to me that you can call these functions with strings, which is very much _not_ something you can do with these functions.

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,6 +96,210 @@ describe('Sayt Component', () => {
...
+    });
+
+    it('should hide SAYT if the event target is nowhere relevant', () => {
+      stub(sayt, 'contains').returns(false);
+      stub(sayt, 'nodeInSearchBar').returns(false);
+      stub(sayt, 'hideSayt');
+
+      sayt.processClick(event);
+
+      expect(sayt.contains).to.be.calledWith(node);
```
jmbtutor: Going back to the earlier discussion, this is actually an instance where we want to make assertions on `node` being the actual object since it's important that we're making assertions on the node itself.

```suggestion
      expect(sayt.contains).to.be.calledWith(sinon.match(node));
```

You will have to import `sinon` from the utils.

The same comment applies to all `contains` assertions below.

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,6 +96,210 @@ describe('Sayt Component', () => {
...
+    });
+
+    it('should prevent the default event action', () => {
+      const preventDefault = spy();
+      const event = { preventDefault };
+      stub(sayt, 'hideSayt');
+
+      sayt.clickCloseSayt(event);
+
+      expect(preventDefault).to.be.calledOnce;
```
jmbtutor: Is it overtesting to assert that it was called exactly once? If I'm not mistaken, the result is the same if we call `preventDefault` more than once.

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,6 +96,210 @@ describe('Sayt Component', () => {
...
+    });
+  });
+
+  describe('nodeInSearchBar()', () => {
+    it('should return true if given node is contained in the search bar', () => {
+      const searchbox = {
+        contains: spy(() => true),
+      };
+      const getElementById = stub(document, 'getElementById').callsFake(() => searchbox);
```
jmbtutor:
```suggestion
      const getElementById = stub(document, 'getElementById').returns(searchbox);
```

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,6 +96,210 @@ describe('Sayt Component', () => {
...
+      expect(getElementById).to.be.calledWith('searchbox-id');
+      expect(searchbox.contains).to.be.calledWith('node');
+      expect(result).to.be.true;
+    });
+
+    it('should return false if given node is not contained in the search bar', () => {
+      const searchbox = {
+        contains: stub().returns(false),
+      };
+      const getElementById = stub(document, 'getElementById').callsFake(() => searchbox);
```
jmbtutor:
```suggestion
      const getElementById = stub(document, 'getElementById').returns(searchbox);
```

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -88,11 +87,11 @@ export default class SearchBox extends Base {
...
    */
   handleInput(e: KeyboardEvent) {
-    this.updateSearchTermValue((e.target as HTMLInputElement).value);
+    const value = (e.target as HTMLInputElement).value;
+    this.updateSearchTermValue(value);
     this.dispatchEvent(
-      new CustomEvent(SEARCHBOX_EVENT.SEARCHBOX_CHANGE, {
-        detail: (e.target as HTMLInputElement).value,
-        bubbles: true
+      this.createCustomEvent(SEARCHBOX_EVENT.SEARCHBOX_CHANGE, {
```
jmbtutor: This function call can probably live on one line.

packages/web-components/@sfx/sayt/src/sayt.ts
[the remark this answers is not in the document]
DanielMoniz-GBI: I'll change to `MouseEvent`, but I'm still getting errors without casting.

packages/web-components/@sfx/sayt/stories/index.ts
DanielMoniz-GBI: The backslash causes errors, but option 2 otherwise works.

packages/web-components/@sfx/sayt/test/sayt.test.ts
DanielMoniz-GBI: It does look nicer.

packages/web-components/@sfx/sayt/test/sayt.test.ts
DanielMoniz-GBI: Yes, the pointer argument makes sense. For the record, I disagree with Uncle Bob - unit tests are code samples only of the unit being tested. Integration tests would be an easier choice for his argument.

packages/web-components/@sfx/sayt/test/sayt.test.ts
DanielMoniz-GBI: As far as I can tell, `sinon.match()` does not match on identity:
```
sinon.match(object);
Requires the value to be not null or undefined and have at least the same properties as expectation.
```

Correct me if I'm wrong, but it doesn't sound like we gain anything from this change given that description.

packages/web-components/@sfx/sayt/test/sayt.test.ts
DanielMoniz-GBI: Yes, this is definitely over-testing.

packages/web-components/@sfx/sayt/test/sayt.test.ts
jmbtutor: Oh, you're right, it should be `sinon.match.same()`. Hopefully I haven't been doing `.match` accidentally elsewhere where I should have used `.match.same` instead.

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,6 +96,211 @@ describe('Sayt Component', () => {
...
+      sayt.hideCorrectSayt(event);
+
+      expect(isCorrectSayt).to.be.calledOnceWith(event);
+      expect(hideSayt).to.not.be.called;
+    });
+  });
+
+  describe('isCorrectSayt()', () => {
+    it('should return true if event provides the correct searchbox ID', () => {
+      const searchbox = sayt.searchbox = 'some-searchbox-id';
```
JoeyDeol: See above comments concerning renaming `searchbox` property.

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,6 +96,211 @@ describe('Sayt Component', () => {
...
+      const event = { detail: { searchbox } };
+
+      const result = sayt.isCorrectSayt(event);
+
+      expect(result).to.be.true;
+    });
+
+    it('should return true if event does not provide a searchbox ID', () => {
+      const event = { detail: {} };
+      sayt.searchbox = 'some-searchbox-id';
```
JoeyDeol: Same as above.

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -65,6 +96,211 @@ describe('Sayt Component', () => {
...
+      stub(sayt, 'nodeInSearchbox').returns(false);
+      stub(sayt, 'hideSayt');
+
+      sayt.processClick(event);
+
+      expect(sayt.contains).to.be.calledWith(sinon.match.same(node));
+      expect(sayt.hideSayt).to.be.called;
+    });
+
+    it('should not SAYT if the event target is contained by SAYT', () => {
```
JoeyDeol:
```suggestion
    it('should not hide SAYT if the event target is contained by SAYT', () => {
```

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -133,10 +130,26 @@ export default class SearchBox extends Base {
...
+   * Returns an event with a standard structure.
+   *
+   * @param type The type (or name) of the event to be emitted.
+   * @param detail A payload to be sent with the event.
+   */
+  createCustomEvent(type: string, detail?: object) {
+    return new CustomEvent(type, {
+      detail: {
+        ...detail,
+        searchbox: this.id,
```
JoeyDeol: Consider renaming the `searchbox` property as mentioned in the comments above.

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -133,10 +130,26 @@ export default class SearchBox extends Base {
...
     this.dispatchEvent(searchBoxClickedEvent);
   }
 
+  /**
+   * Returns an event with a standard structure.
+   *
+   * @param type The type (or name) of the event to be emitted.
+   * @param detail A payload to be sent with the event.
+   */
+  createCustomEvent(type: string, detail?: object) {
```
JoeyDeol: Should this have a return type of `CustomEvent`?

packages/web-components/@sfx/sayt/stories/index.ts
```diff
@@ -30,28 +53,85 @@ storiesOf('Components|SAYT', module)
...
+
+${ input2 }
+${ sayt2 }`) }
+    `;
+  }, {
+    notes: {
+      markdown:`
+        #Search As You Type (SAYT)
+        Demonstrating multiple SAYT components.This proves that eachSearch/SAYT
+        pair acts independently.
```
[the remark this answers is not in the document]
jmbtutor: The original "acts" is correct. Although "pair" refers to two members, the word itself is singular.

packages/web-components/@sfx/search-box/src/search-box.ts
jmbtutor: Come to think of it, this would be better typed as the following:

```suggestion
  createCustomEvent<T>(type: string, detail?: T): CustomEvent<T> {
```

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -15,11 +16,33 @@ export default class Sayt extends LitElement {
    * Determines the visibility of the `sayt` component.
    */
   @property({ type: Boolean, reflect: true }) visible = false;
+  /**
+   * Stores the ID of the relevant search element.
+   */
+  @property({ type: String, reflect: true }) searchbox = '';
```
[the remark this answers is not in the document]
jmbtutor: Precedent in HTML says otherwise. See the `form` attribute for the various form elements, `for` for `<label>`, and loosely `usemap` for `` elements.

packages/web-components/@sfx/search-box/src/search-box.ts
DanielMoniz-GBI: True! Good catch to you both.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -64,15 +93,105 @@ export default class Sayt extends LitElement {
...
+  }
+
+  /**
+   * Processes a click event in order to close SAYT under the right conditions.
+   * 
+   * @param event The click event.
+   */
+  processClick(event: MouseEvent) {
+    const target = event.target as Node;
+    if (this.contains(target) || this.nodeInSearchbox(target)) return;
```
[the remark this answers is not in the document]
DanielMoniz-GBI: If I return something here (eg. `false`), I'd have to `return true` at the end. Given that its job is to process a click event, I personally don't think there is much point in returning anything. I'm only returning to prevent future logic from running.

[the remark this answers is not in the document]
jmbtutor: > Perhaps a boolean to indicate that the click as been processed in an expected manner?

If it had been processed in an unexpected manner, it would be better to throw an exception instead of relying on a return code.

PR #16, opened 2019-08-06
SFX-250 Integrate View and Logic Layer 

ssbelle:
- Pull in logic layer through script tag pointing to CDN
- Add search box, sayt, and a products list 
- Init core and all plugins
- Register all plugins with core
- Listen for search response and add those items to a temp product grid
- Temporary hardcoded autocomplete list


demo/demo.html
```diff
@@ -0,0 +1,40 @@
...
+      const searchDriverPlugin = new sfxPlugins.SearchDriverPlugin();
+      // register all plugins with core
+      core.register([saytPlugin, saytDriverPlugin, domEventsPlugin, searchPlugin, searchDriverPlugin]);
+
+      // search integration/functionality test
+      window.addEventListener("sfx::search_response", (e) => {
+        const testListResults = e.detail.records.map(item => {
+          return { label : item.title };
+        });
+        document.getElementsByClassName('product-grid')[0].setAttribute('items', JSON.stringify(testListResults));
```
DanielMoniz-GBI: Can we split this up over two lines?

PR #19, opened 2019-08-12
SFX-299: Fix container size of components

DanielMoniz-GBI:
This adds `display: block` to most components. Note that in any component that can be shown/hidden, the `block` display must be active only when shown.

Caveats:
* The following components do not have a correct container size when used in isolation:
  * `List`
  * `Product`
  * `Base`
  (None of the above components are currently used in isolation, so this is technically not yet a problem. Additionally, Base should never be used in isolation.)

### DIFF
--- a/packages/web-components/@sfx/products/src/products.ts
+++ b/packages/web-components/@sfx/products/src/products.ts
@@ -69,6 +69,10 @@ export default class Products extends LitElement {
           flex-wrap: wrap;
         }
 
+        sfx-products[hidden] {
+          display: none;
+        }
+
         sfx-product {
           display: block;
         }
--- a/packages/web-components/@sfx/products/src/products.ts
+++ b/packages/web-components/@sfx/products/src/products.ts
@@ -72,6 +72,10 @@ export default class Products extends LitElement {
         sfx-product {
           display: block;
         }
+
+        sfx-product[hidden] {
+          display: none;
+        }
       </style>
 
       ${this.products.map(product => {
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -191,6 +191,10 @@ export default class Sayt extends LitElement {
           sfx-sayt {
             display: block;
           }
+
+          sfx-sayt[hidden] {
+            display: none;
+          }
         ` : ''}
       </style>
       ${ this.showCloseButton ?
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -187,15 +187,13 @@ export default class Sayt extends LitElement {
   render() {
     return html`
       <style>
-        ${ this.visible ? `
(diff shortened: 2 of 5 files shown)


packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -186,6 +186,13 @@ export default class Sayt extends LitElement {
    */
   render() {
     return html`
+      <style>
+        ${ this.visible ? `
+          sfx-sayt {
+            display: block;
+          }
+        ` : ''}
+      </style>
```
DanielMoniz-GBI: Note that this does create empty `<style>` tags when invisible.

DanielMoniz-GBI:
@ChantalDesrochers My impression is that, given that we're still in the "unreleased" category, we don't really need to add changes. If a component splits into two, or something else major, we can consider recording that. Until then, we are just 'adding' the relevant components. Later we'll be officially changing them.

packages/web-components/@sfx/sayt/src/sayt.ts
jmbtutor: If there happens to be multiple `sfx-sayt` components on the page, this is not going to work. What would be better is to define two static styles, one for visible and one for hidden:

```css
sfx-sayt {
  display: none;
}

sfx-sayt[visible] {
  display: block;
}
```

Or the other way:
```css
sfx-sayt {
  display: block;
}

sfx-sayt[hidden] {
  display: none;
}
```

The browser should already have styles for `[hidden]`, so I wonder if simply defining `sfx-sayt` to be `display: block` is enough.

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -170,6 +170,11 @@ export default class SearchBox extends Base {
 
   render() {
     return html`
+      <style>
+        sfx-search-box {
+          display: block;
```
jmbtutor: I think search box should be `inline-block`. One should be able to display this component alongside some text. For the record, the computed display styles of an input box and a button are "inline" (Firefox)/"inline-block" (Chrome) and "inline-block" respectively

packages/web-components/@sfx/sayt/src/sayt.ts
DanielMoniz-GBI: Good catch! I didn't notice it, despite it being apparent in a Storybook story. Will change.

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -69,9 +69,17 @@ export default class Products extends LitElement {
           flex-wrap: wrap;
         }
 
+        sfx-products[hidden] {
+          display: none;
+        }
+
         sfx-product {
```
jmbtutor: Oh, this is where this style ended up. This should live in the `Product` component. That said, if we want to push that off to the future base styles ticket, we can do so as well.

jmbtutor: We'll deal with this in a future ticket.

PR #22, opened 2019-08-16
SFX-295: Sayt wc and sayt driver connections complete

ssbelle:
- Multiple sayts function independently on demo page
- Temporarily pointing to a bundled logic file until logic PR is approved and CDN updated
- Added second sayt to page with its own searchbox
- Removed hardcoded autocomplete call as that functionality is now hooked up
- Added listener in sayt wrapper for searchbox change
- Add attribute to customize a min char in sayt or default to 3
- On searchbox change if sayt is correct, and min char is correct,  trigger sayt request
- Send searchbox to sayt driver 

NOTE: 
- Tests are WIP


packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -120,6 +121,12 @@ describe('Sayt Component', () => {
     });
   });
 
+  describe('requestSayt()', () => {
+    it('should return if the event does not provide the correct searchbox ID', () => {
+
```
ssbelle: please not tests are a WIP as I wait for other feedback to be made, thanks

demo/demo.html
```diff
@@ -3,11 +3,15 @@
   <head>
     <title>Demo of Logic and View Integration</title>
     <script src="../dist/components.js"></script>
-    <script src="https://cdn.groupbycloud.com/sfx-all-plugins-bundle.js"></script>
+    <script src='./temp/sfx-all-plugins-bundle.js'></script> 
+    <!-- above tag to be removed when the cdn PR passes with the updated bundle -->
```
ssbelle: please note that this script tag is temporary to enable you all to run my demo  while I wait for my logic pr to pass and the CDN to be updated. Thanks!

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -123,6 +129,23 @@ export default class Sayt extends LitElement {
...
 
+  /**
+   * Determines whether an autocomplete call should be made based on minchars attribute.
+   * Then fires the autocomplete fetch data event to communicate with the sayt driver.
+   *
+   * @param event An event that can contain a searchbox ID.
+   */
+  requestSayt(event: CustomEvent) {
+    if (!this.isCorrectSayt(event)) return;
+    if(event.detail.value.length >= this.minchars) {
```
DanielMoniz-GBI: You can take the same strategy as the previous line to avoid having to indent a bunch of code. Basically, check if there are _not enough_ characters, and just return if that's the case.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -123,6 +129,23 @@ export default class Sayt extends LitElement {
...
+      const requestSaytResults = new CustomEvent('sfx::autocomplete_fetch_data', {
+        detail: { query: event.detail.value, searchbox: event.detail.searchbox },
+        bubbles: true,
+      });
+    window.dispatchEvent(requestSaytResults);
```
DanielMoniz-GBI: The indentation is a bit off here.

packages/web-components/@sfx/sayt/test/sayt.test.ts
```diff
@@ -22,6 +22,7 @@ describe('Sayt Component', () => {
       expect(addEventListener).to.be.calledWith('click', sayt.processClick);
       expect(addEventListener).to.be.calledWith('keydown', sayt.processKeyEvent);
       expect(addEventListener).to.be.calledWith(AUTOCOMPLETE_RECEIVED_RESULTS_EVENT, sayt.showCorrectSayt);
+      expect(addEventListener).to.be.calledWith('sfx::searchbox_change', this.requestSayt);
```
DanielMoniz-GBI: In case we miss this later when you are adding more tests: we should also be asserting that `this.requestSayt` is removed in the `disconnectedCallback()` by adding to the test below.

packages/web-components/@sfx/sayt/CHANGELOG.md
```diff
@@ -8,3 +8,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - SFX-191: Added the `sayt` component.
   - This component renders `sayt` related components such as `sfx-autocomplete` and `sfx-products`.
+  -SFX-295: Added functionality to accommodate multiple `sayts`.
```
DanielMoniz-GBI: We don't need to add this line. Given that we're still in an unreleased state, we can largely ignore the changelog. If there is a major change that is not core to the idea of SAYT, it could be worth adding for clarity.

demo/demo.html
jmbtutor: You might find `git add -p` to be useful in the future. It allows you to commit only parts of files.

packages/web-components/@sfx/sayt/src/sayt.ts
geoffcoutts: Lines 139 and 140 could be combined:
```suggestion
    if (!this.isCorrectSayt(event) || !event.detail.value.length >= this.minchars) return;
```

And then you don't need the second if statement.

demo/demo.html
```diff
@@ -3,11 +3,15 @@
...
+    <!-- <script src="https://cdn.groupbycloud.com/sfx-all-plugins-bundle.js"></script> -->
   </head>
   <body>
-    <sfx-search-box placeholder="search" clearbutton searchbutton></sfx-search-box>
-    <sfx-sayt visible></sfx-sayt>
+    <sfx-search-box placeholder="search" clearbutton searchbutton id='searchbox1'></sfx-search-box>
+    <sfx-sayt title='basic' searchbox='searchbox1'></sfx-sayt>
```
geoffcoutts: Is there any value in using the title attribute for the two `sfx-sayt` components? Also, does `minch` have any value?

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -123,6 +129,23 @@ export default class Sayt extends LitElement {
     }
   }
 
+  /**
+   * Determines whether an autocomplete call should be made based on minchars attribute.
+   * Then fires the autocomplete fetch data event to communicate with the sayt driver.
```
jmbtutor: Given that the View layer is independent, it shouldn't assume the presence of any drivers, which means that we shouldn't mention the drivers in the documentation (or any thing in the Logic layer, for that matter).

Your description should also not be describing what the function does in a step-by-step manner. In most cases, it can help to treat the function as a black box: you feed it some input and [something] happens. Remember that the first sentence is the most important and should be able to stand alone.

I would word it something like this (wrapped properly):

> Dispatches an `sfx::autocomplete_fetch_data` event with the provided data. The event will only be dispatched if this SAYT component is the "correct" SAYT (see [[isCorrectSayt]]) and if the term is at least [[minchars]] long.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -123,6 +129,23 @@ export default class Sayt extends LitElement {
...
+   *
+   * @param event An event that can contain a searchbox ID.
```
jmbtutor: I realize that this is in the other functions, but this is much less descriptive in this context. Try to describe what this event is; it might help to think about why this function needs this event.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -28,7 +28,10 @@ export default class Sayt extends LitElement {
    * Shows a button to allow for closing SAYT manually.
    */
   @property({ type: Boolean, reflect: true }) showCloseButton = false;
-
+  /**
+   * Customizes the number of characters entered in searchbox before opening SAYT.
+   */
+  @property({ type: Number, reflect: true }) minchars = 3;
```
jmbtutor: I don't like this name. "minchars" of what? Seeing this on a SAYT tag says to me that something displayed in the SAYT component should have at minimum 3 characters. At least, that's the case for me. I think "min" and "length" should be in the name, but "minlength" has the same problem. @geoffcoutts suggested "minSearchLength" or "minTermLength"; it's still not great, but it's better.

As for why you shouldn't use "chars", a "character" can be tricky to define. Take the following two e's: é and é. They look identical and like single characters, but you'll find that the two actually have different lengths:

```js
console.log('é'.length); // 1
console.log('é'.length); // 2
```

This is because the first one is a combined character, whereas the second is a character with a combining diacritic. "Length" is more predictable as it tends to deal with the number of code points.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -28,7 +28,10 @@ export default class Sayt extends LitElement {
    * Shows a button to allow for closing SAYT manually.
    */
   @property({ type: Boolean, reflect: true }) showCloseButton = false;
-
+  /**
+   * Customizes the number of characters entered in searchbox before opening SAYT.
```
jmbtutor: Remember that it's also valid to read from this property, so the description should make sense in that context as well. Maybe something like this?

```suggestion
   * The minimum length of the search term required before a SAYT request will be made with it.
```

demo/demo.html
```diff
@@ -3,11 +3,15 @@
...
+    <sfx-search-box placeholder="search" clearbutton searchbutton id='searchbox2'></sfx-search-box>
+    <sfx-sayt minchars=5 title='minch' searchbox='searchbox2'></sfx-sayt>
```
jmbtutor: What does "minch" mean?

demo/demo.html
```diff
@@ -3,11 +3,15 @@
...
+    <script src='./temp/sfx-all-plugins-bundle.js'></script> 
+    <!-- above tag to be removed when the cdn PR passes with the updated bundle -->
+    <!-- <script src="https://cdn.groupbycloud.com/sfx-all-plugins-bundle.js"></script> -->
   </head>
   <body>
-    <sfx-search-box placeholder="search" clearbutton searchbutton></sfx-search-box>
-    <sfx-sayt visible></sfx-sayt>
+    <sfx-search-box placeholder="search" clearbutton searchbutton id='searchbox1'></sfx-search-box>
+    <sfx-sayt title='basic' searchbox='searchbox1'></sfx-sayt>
+    <sfx-search-box placeholder="search" clearbutton searchbutton id='searchbox2'></sfx-search-box>
```
jmbtutor: I would put `id` closer to the front for visibility.

demo/demo.html
```diff
@@ -3,11 +3,15 @@
...
     <script src="../dist/components.js"></script>
-    <script src="https://cdn.groupbycloud.com/sfx-all-plugins-bundle.js"></script>
+    <script src='./temp/sfx-all-plugins-bundle.js'></script> 
+    <!-- above tag to be removed when the cdn PR passes with the updated bundle -->
+    <!-- <script src="https://cdn.groupbycloud.com/sfx-all-plugins-bundle.js"></script> -->
   </head>
   <body>
-    <sfx-search-box placeholder="search" clearbutton searchbutton></sfx-search-box>
-    <sfx-sayt visible></sfx-sayt>
+    <sfx-search-box placeholder="search" clearbutton searchbutton id='searchbox1'></sfx-search-box>
```
jmbtutor: I would put `id` near the beginning for visibility. I would also keep the boolean attributes at the end.

demo/demo.html
```diff
@@ -3,20 +3,24 @@
...
+    <!-- above tag to be removed when the cdn PR passes with the updated bundle -->
+    <!-- <script src="https://cdn.groupbycloud.com/sfx-all-plugins-bundle.js"></script> -->
   </head>
   <body>
-    <sfx-search-box placeholder="search" clearbutton searchbutton></sfx-search-box>
-    <sfx-sayt visible></sfx-sayt>
+    <sfx-search-box id="searchbox1" placeholder="search" clearbutton searchbutton></sfx-search-box>
+    <sfx-sayt searchbox="searchbox1"></sfx-sayt>
+    <sfx-search-box id="searchbox2" placeholder="search" clearbutton searchbutton></sfx-search-box>
+    <sfx-sayt searchbox="searchbox2" minSearchLength=5></sfx-sayt>
```
jmbtutor: HTML attributes should be all lowercase. Also, even if they're numbers, attribute values should still have quotes around them.

```suggestion
    <sfx-sayt searchbox="searchbox2" minsearchlength="5"></sfx-sayt>
```

demo/demo.html
```diff
@@ -3,20 +3,24 @@
...
     <sfx-list class="product-grid"></sfx-list>
 
     <script>
       // init core and plugins
       const core = new sfxCore.Core();
       const domEventsPlugin = new sfxPlugins.DomEventsPlugin()
-      const saytPlugin = new sfxPlugins.SaytPlugin({subdomain: 'apparel', collection: 'products'});
+      const saytPlugin = new sfxPlugins.SaytPlugin({subdomain: "apparel", collection: "products"});
```
jmbtutor: We can keep the single quotes in the JavaScript portions.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -130,6 +136,23 @@ export default class Sayt extends LitElement {
     }
   }
 
+  /**
+   * Dispatches an `sfx::autocomplete_fetch_data` event with the provided data.
+   * The event will only be dispatched if this SAYT component is the "correct" SAYT
+   * (see [[isCorrectSayt]]) and if the term is at least [[minSearchLength]] long.
+   *
+   * @param event An event that can contain a searchbox ID to identify its associated sayt.
```
jmbtutor: There's more than just the searchbox ID in this event, though. It also includes the search term. Perhaps something like this:

```suggestion
   * @param event The searchbox change event dispatched from the searchbox.
```

demo/demo.html
```diff
@@ -3,20 +3,24 @@
...
+    <sfx-sayt searchbox="searchbox1"></sfx-sayt>
+    <sfx-search-box id="searchbox2" placeholder="search" clearbutton searchbutton></sfx-search-box>
+    <sfx-sayt searchbox="searchbox2" minSearchLength=5></sfx-sayt>
     <sfx-list class="product-grid"></sfx-list>
 
     <script>
       // init core and plugins
       const core = new sfxCore.Core();
       const domEventsPlugin = new sfxPlugins.DomEventsPlugin()
-      const saytPlugin = new sfxPlugins.SaytPlugin({subdomain: 'apparel', collection: 'products'});
```
geoffcoutts: `"` should be used for tag attributes, however, script tags should use `'` for javascript.

demo/demo.html
geoffcoutts: Please use spaces inside of `{...}` in JS. 
```suggestion
      const saytPlugin = new sfxPlugins.SaytPlugin({ subdomain: "apparel", collection: "products" });
```

demo/demo.html
```diff
@@ -3,20 +3,24 @@
...
+      const saytPlugin = new sfxPlugins.SaytPlugin({subdomain: "apparel", collection: "products"});
       const saytDriverPlugin = new sfxPlugins.SaytDriverPlugin();
-      const searchPlugin = new sfxPlugins.SearchPlugin({ customerId: 'apparel', collection: 'products', area: 'Storefront'});
+      const searchPlugin = new sfxPlugins.SearchPlugin({ customerId: "apparel", collection: "products", area: "Storefront"});
```
geoffcoutts: Same as above
```suggestion
      const searchPlugin = new sfxPlugins.SearchPlugin({ customerId: "apparel", collection: "products", area: "Storefront" });
```

demo/demo.html
```diff
@@ -26,16 +30,9 @@
         const testListResults = e.detail.records.map(item => {
           return { label : item.title };
         });
-        document.getElementsByClassName('product-grid')[0]
-        .setAttribute('items', JSON.stringify(testListResults));
+        document.getElementsByClassName("product-grid")[0]
```
geoffcoutts: Single quotes for JS, please. This line and the next.

packages/web-components/@sfx/autocomplete/test/unit/autocomplete.test.ts
```diff
@@ -80,14 +80,15 @@ describe('Autcomplete Component', () => {
 
   describe('receivedResults', () => {
     it('should update the results property in response to data received', () => {
-      const detail = [
+
```
jmbtutor: Remove extra line.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -2,7 +2,7 @@ import { LitElement, customElement, html, property, PropertyValues } from 'lit-e
 import { PRODUCTS_EVENT } from '@sfx/products';
 import { SAYT_EVENT } from './events';
 import { AUTOCOMPLETE_RECEIVED_RESULTS_EVENT } from '../../autocomplete/src/events';
-
+import { SEARCHBOX_EVENT } from '../../search-box/src/events';
 /**
```
jmbtutor: Add a blank line above to separate imports from the rest of the file.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -91,6 +109,36 @@ export default class Sayt extends LitElement {
...
+
+  /**
+   * Toggle the events being registered and unregisterd when the `searchbox` property changes.
+   *
+   * @param toggle A string to indicate the type of eventListener(add or remove).
+   * @param eventName The name of the eventListener to register or unregister.
+   * @param eventCallback The name of the callback function triggered by the eventListener.
+   * @param searchboxType The `oldSearchbox` or `this.searchbox`.
+   */
+  toggleSearchboxListener(toggle: string, eventName: string, eventCallback: any, searchboxType?: string) :void {
```
jmbtutor: I think a better factoring is to simply pass down the searchbox ID (type `string`) and whether to add or remove (type `'add' | 'remove'`). You already know which event callback to use and under which event name based on the searchbox ID: if it's on  the searchbox itself, it's `input` and `this.processSearchboxInput`, and if it's on `window`, it's `sfx::searchbox_change` and `this.processSfxSearchboxChange`. If you do it this way, you can replace all the `if` statements in the calling code with a simple call to this function.

I also wouldn't call it "toggle" because you aren't really toggling anything in this function. Maybe `setSearchboxListener`.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -91,6 +109,36 @@ export default class Sayt extends LitElement {
...
+    if (searchboxType) {
+      const searchbox :any = document.getElementById(searchboxType);
```
jmbtutor:
```suggestion
      const searchbox = document.getElementById(searchboxType) as HTMLInputElement;
```

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -91,6 +109,36 @@ export default class Sayt extends LitElement {
...
+      if (searchbox) searchbox[`${toggle}EventListener`](eventName, eventCallback);
+    } else {
+      (window as any)[`${toggle}EventListener`](eventName, eventCallback);
```
jmbtutor: I don't think `window` needs to be cast here.

```suggestion
      window[`${toggle}EventListener`](eventName, eventCallback);
```

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -58,16 +91,113 @@ describe('Sayt Component', () => {
...
+    it('should unregister listener for searchbox change if searchbox is not given', () => {
+      const removeEventListener = stub(window, 'removeEventListener');
+      sayt.searchbox = '';
+
+      sayt.disconnectedCallback();
+
+      expect(removeEventListener).to.be.calledWith('sfx::searchbox_change', sayt.processSfxSearchboxChange);
+    });
+
+    it('should not unregister listener for input change if specified searchbox element exists', () => {
```
jmbtutor: Your tests are going to change and some of these will probably go away. If you implement the suggested changes about `toggleSearchboxListener`, your test will simply be asserting that the function is called.

demo/demo.html
ssbelle: @DanielMoniz-GBI I originally had it camelcase here, and camelcase in the sayt.ts file that you commented on. @jmbtutor requested this be changed to all lowercase as you can see above, and having it as `min-search-length` in the html and receiving it as camel case `minSearchLength` in sayt.ts was not working? @geoffcoutts maybe something to 100% solve when we pair

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -33,7 +33,10 @@ export default class Sayt extends LitElement {
    * Shows a button to allow for closing SAYT manually.
    */
   @property({ type: Boolean, reflect: true }) showCloseButton = false;
-
+  /**
+   * The minimum length of the search term required before a SAYT request will be made with it.
+   */
+  @property({ type: Number, reflect: true }) minsearchlength = 3;
```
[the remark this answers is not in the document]
geoffcoutts: @DanielMoniz-GBI, @ssbelle 
Case conversion from HTML attributes isn't automatically handled by lit-element and would need to be done manually. Please see https://github.com/Polymer/lit-element/issues/29

demo/demo.html
ssbelle: nvm^

demo/demo.html
```diff
@@ -20,22 +20,14 @@
       const searchDriverPlugin = new sfxPlugins.SearchDriverPlugin();
       // register all plugins with core
       core.register([saytPlugin, saytDriverPlugin, domEventsPlugin, searchPlugin, searchDriverPlugin]);
-
       // search integration/functionality test
-      window.addEventListener("sfx::search_response", (e) => {
+      window.addEventListener('sfx::search_response', (e) => {
```
ssbelle: reverted demo.html to old code for Joeys demo branch. Minus the unneeded autocomplete call and the one quotes change

PR #31, opened 2019-09-03
Sfx 333 squash

DanielMoniz-GBI:


packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -222,6 +229,120 @@ describe('Sayt Component', () => {
     });
   });
 
+  describe('requestSayt()', () => {
+    let dispatchEvent, query, searchbox;
+    beforeEach(() => {
+      query = 'some-query';
```
jmbtutor: This default is only used in one test, so I think we can remove `query` as a shared variable and use it only in the tests.

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -222,6 +229,120 @@ describe('Sayt Component', () => {
...
+      searchbox = 'some-searchbox-id';
```
jmbtutor: Since this doesn't change in the tests, this should just be a `const` outside the `beforeEach`.

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -222,6 +229,120 @@ describe('Sayt Component', () => {
...
+    });
+  });
+
+  describe('requestSaytAutocompleteTerms', () => {
+    let dispatchEvent, query, searchbox;
+    beforeEach(() => {
+      query = 'some-query';
+      searchbox = 'some-searchbox-id';
+      dispatchEvent = stub(window, 'dispatchEvent');
+    });
```
jmbtutor: Same comments as the `requestSayt()` `beforeEach`.

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -222,6 +229,120 @@ describe('Sayt Component', () => {
     });
   });
 
+  describe('requestSayt()', () => {
```
jmbtutor: Be consistent. Either have `()` on all or none.

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -222,6 +229,120 @@ describe('Sayt Component', () => {
...
+      const eventObj = {};
+      const CustomEvent = stub(window, 'CustomEvent').returns(eventObj);
+
+      sayt.requestSaytAutocompleteTerms(query, searchbox);
+
+      expect(CustomEvent).to.be.calledWith('sfx::autocomplete_fetch_data', {
+        detail: { query, searchbox },
+        bubbles: true,
+      });
+      expect(dispatchEvent).to.be.calledWith(eventObj);
```
jmbtutor: This line does not check for an exact reference match. If you want an exact reference match, use `sinon.match.same(eventOjb)`. Whether you do or do not care about the reference match, you should add some content inside so that if the test fails, it's easy to identify. Same thing goes for similar tests.

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -222,6 +229,120 @@ describe('Sayt Component', () => {
...
+      expect(CustomEvent).to.be.calledWith('sfx::sayt_products_request', {
+        detail: { query, searchbox: undefined },
+        bubbles: true,
+      });
+      expect(dispatchEvent).to.be.calledWith(eventObj);
+    });
+  });
+
+  // @TODO Add tests for processSearchboxInput()
+  // @TODO Add tests for processSfxSearchboxChange()
```
jmbtutor: You can add empty `describe` and `it` blocks; they'll show up as skipped.

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -184,3 +196,7 @@ export default class SearchBox extends Base {
     `;
   }
 }
+export interface SearchConfig {
+  area: string | undefined,
+  collection: string | undefined,
```
jmbtutor: Is it better to make these optional instead?

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -222,6 +229,120 @@ describe('Sayt Component', () => {
...
+      dispatchEvent = stub(window, 'dispatchEvent');
+    });
+
+    it('should dispatch an event with a payload that includes the query and searchbox', () => {
+      const eventObj = {};
+      const CustomEvent = stub(window, 'CustomEvent').returns(eventObj);
+
+      sayt.requestSaytAutocompleteTerms(query, searchbox);
+
+      expect(CustomEvent).to.be.calledWith('sfx::autocomplete_fetch_data', {
```
geoffcoutts: Tests for `requestSaytAutocompleteTerms()` and `requestSaytProducts()` should reference the correct exported events.

packages/web-components/@sfx/search-box/src/search-box.ts
```diff
@@ -184,3 +196,7 @@ export default class SearchBox extends Base {
     `;
   }
 }
+export interface SearchConfig {
```
geoffcoutts: This interface isn't actually being used. Is the intention to use it as part of `emitSearchEvent()`?

packages/web-components/@sfx/autocomplete/src/autocomplete.ts
```diff
@@ -50,7 +57,24 @@ export default class Autocomplete extends LitElement {
...
+  }
+
+  /**
+   * Dispatches an `sfx::sayt_hover_autocomplete_term` event with the Sayt autocomplete term.
+   *
+   * @param event A MouseEvent that contains a Sayt autocomplete term.
+   */
+  handleHoverTerm(event: MouseEvent) {
+    const target = event.target as HTMLElement;
+    if (target.tagName !== 'LI') return;
```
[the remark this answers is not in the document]
DanielMoniz-GBI: Ah, I didn't catch the exception. I thought it capitalized all tag names, in which case forcing an additional case change would be unnecessary.

packages/web-components/@sfx/autocomplete/src/autocomplete.ts
```diff
@@ -50,7 +57,24 @@ export default class Autocomplete extends LitElement {
...
+    const term = target.innerText;
+    const sentEvent = new CustomEvent(HOVER_AUTOCOMPLETE_TERM_EVENT, {
```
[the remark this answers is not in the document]
DanielMoniz-GBI: It's to distinguish it from the received event, which is defined as a parameter. Do you have a suggestion as to the naming of the two?

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
JoeyDeol: Please resolve these TODOs.

packages/web-components/@sfx/autocomplete/test/unit/autocomplete.test.ts
```diff
@@ -88,5 +98,50 @@ describe('Autcomplete Component', () => {
 
       expect(autocomplete.results).to.deep.equal(results);
     });
+
+    it('should set the results property to an empty array if the response data is undefined', () => {
```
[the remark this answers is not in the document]
DanielMoniz-GBI: @geoffcoutts Can you elaborate on this? Do you mean if the event payload contains an error because the autocomplete call failed? This test ensures that the products are set to an array even if there are no results, which should handle any kind of response that does not include results. Let me know.

packages/web-components/@sfx/products/src/products.ts
```diff
@@ -8,7 +8,9 @@ import {
 import { ProductModel } from '@sfx/product';
 
 /** The name of the event that contains product data. */
-export const PRODUCTS_EVENT: string = 'sfx::provide_products'
+export const PRODUCTS_RESPONSE_EVENT: string = 'sfx::sayt_products_response';
+/** The name of the event that contains product request. */
+export const PRODUCTS_REQUEST_EVENT: string = 'sfx::sayt_products_request';
```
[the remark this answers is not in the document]
DanielMoniz-GBI: Yeah, the location of this could be debated. Resolving for now because of the events repository, as you mentioned.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -174,13 +185,49 @@ export default class Sayt extends LitElement {
...
+  }
+
+  /**
+   * Handles how the hover on Sayt autocomplete terms updates the sayt products.
+   * Triggers a request of Sayt products using the query and searchbox data.
+   *
+   * @param event The hover event dispatched from autocomplete.
+   */
+  handleAutocompleteTermHover(event: CustomEvent) {
+    const query = event.detail.query;
```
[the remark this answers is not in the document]
DanielMoniz-GBI: @jmbtutor I'll inline it, but I'm curious why you prefer destructuring here.

packages/web-components/@sfx/autocomplete/test/unit/autocomplete.test.ts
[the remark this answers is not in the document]
geoffcoutts: @DanielMoniz-GBI  True, it would set it to blank regardless. I'm guessing any errors that would potentially be passed to this function from search calls would also show up as either a failed call in the network tab or in the search call's response?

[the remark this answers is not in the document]
DanielMoniz-GBI: Yeah, I think any errors would have to be dealt with in the logic layer. If it's making it this far, we're assuming it's arriving in a certain format.

We also have a separate event for error responses, I believe, so we can assume that this event contains the results from a successful search.

packages/web-components/@sfx/autocomplete/src/autocomplete.ts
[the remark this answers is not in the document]
geoffcoutts: Okay, I see why you named them how you did. You can keep them as is, the names just feel weird.

packages/web-components/@sfx/autocomplete/test/unit/autocomplete.test.ts
```diff
@@ -88,5 +98,50 @@ describe('Autcomplete Component', () => {
...
+      autocomplete.handleHoverTerm(mouseEvent);
+
+      expect(dispatchEvent).to.not.be.called;
+    });
+
+    it('should emit an event when hovering an autocomplete term', () => {
+      CustomEvent = stub(window, 'CustomEvent').returns({});
+      const mouseEvent = {
+        target: {
+          tagName: 'LI',
```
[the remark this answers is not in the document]
DanielMoniz-GBI: The key piece here is that we're only listening to events on the element itself. This function is indeed built to react to any li being hovered, but it should only be called when a child of autocomplete is hovered.

packages/web-components/@sfx/autocomplete/src/autocomplete.ts
```diff
@@ -34,6 +39,7 @@ export default class Autocomplete extends LitElement {
   connectedCallback() {
     super.connectedCallback();
     window.addEventListener(AUTOCOMPLETE_RECEIVED_RESULTS_EVENT, this.receivedResults);
+    this.addEventListener('mouseover', this.handleHoverTerm);
```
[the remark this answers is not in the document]
DanielMoniz-GBI: Absolutely. Good catch.

packages/web-components/@sfx/autocomplete/src/autocomplete.ts
```diff
@@ -2,11 +2,15 @@ import { customElement, html, property, LitElement } from 'lit-element';
 import '@sfx/ui';
 
 // TODO: replace with @sfx/events import - added temporarily for import in storybook/common
+// event descriptions to be contained in events repo
 export const AUTOCOMPLETE_RECEIVED_RESULTS_EVENT = 'sfx::autocomplete_received_results';
+export const HOVER_AUTOCOMPLETE_TERM_EVENT = 'sfx::sayt_hover_autocomplete_term';
```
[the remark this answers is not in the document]
DanielMoniz-GBI: I'm going to leave this one, but bump @jmbtutor to take note of it when updating the Events repo. That will be the easiest place to make these kinds of changes.

packages/web-components/@sfx/products/test/unit/products.test.ts
```diff
@@ -1,5 +1,5 @@
 import { expect, sinon, stub } from '../utils';
-import Products, { PRODUCTS_EVENT } from '../../src/products';
+import { Products, PRODUCTS_RESPONSE_EVENT } from '@sfx/products';
```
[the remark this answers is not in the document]
DanielMoniz-GBI: I have some memory of a suggestion moving us toward doing this, but I suppose we never agreed on it.

I get the part about making local imports clear. Overall, though, are you suggesting we import from local files more generally, or that we only do it in test files? Please just make your actual suggestion clearer, as most of what you said is stated in terms of negatives. I'm having a hard time interpreting what your monorepo argument without knowing the scope of your argument (tests vs global).

packages/web-components/@sfx/autocomplete/test/unit/autocomplete.test.ts
```diff
@@ -28,6 +29,14 @@ describe('Autcomplete Component', () => {
   });
 
   describe('connectedCallback', () => {
+    let windowAddEventListener,
+        autocompleteAddEventListener;
```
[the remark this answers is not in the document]
DanielMoniz-GBI: Missed this. I tried to convert them all back.

packages/web-components/@sfx/autocomplete/src/autocomplete.ts
```diff
@@ -26,6 +30,7 @@ export default class Autocomplete extends LitElement {
   constructor() {
     super();
     this.receivedResults = this.receivedResults.bind(this);
+    this.handleHoverTerm = this.handleHoverTerm.bind(this);
```
geoffcoutts: This no longer needs to be bound here if using `@` in the [render method](https://github.com/groupby/sfx-view/pull/31/files#diff-15c2226c616cd93d909210ac1585dd59R101)

packages/web-components/@sfx/products/test/unit/products.test.ts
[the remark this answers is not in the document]
jmbtutor: I'm suggesting we revert back to `'../../src/products'`. In the context of importing files from the same package, we should be using relative local imports. Between packages should be done as a non-local import (e.g. tests in `@sfx/sayt` should import from `@sfx/products`, not `../../../../products` or whatever the path may be).

Same goes for `autocomplete.test.ts:4`.

packages/web-components/@sfx/autocomplete/test/unit/autocomplete.test.ts
```diff
@@ -28,6 +29,14 @@ describe('Autcomplete Component', () => {
   });
 
   describe('connectedCallback', () => {
+    let windowAddEventListener;
+    let autocompleteAddEventListener;
+
+    beforeEach(() => {
+      windowAddEventListener = stub(window, 'addEventListener');
+      autocompleteAddEventListener = stub(autocomplete, 'addEventListener');
```
geoffcoutts: This stub isn't being used.

packages/web-components/@sfx/autocomplete/test/unit/autocomplete.test.ts
```diff
@@ -36,19 +45,25 @@ describe('Autcomplete Component', () => {
...
     });
   });
 
   describe('disconnectedCallback', () => {
+    let windowRemoveEventListener;
+    let autocompleteRemoveEventListener;
+
+    beforeEach(() => {
+      windowRemoveEventListener = stub(window, 'removeEventListener');
+      autocompleteRemoveEventListener = stub(autocomplete, 'removeEventListener');
```
geoffcoutts: This stub isn't being used.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -174,11 +183,54 @@ export default class Sayt extends LitElement {
...
+   *
+   * @param query The search term to use.
+   * @param searchbox The optional searchbox ID associated with this search.
+   */
+  requestSaytAutocompleteTerms(query: string, searchbox?: string) {
+    this.dispatchRequestEvent(AUTOCOMPLETE_REQUEST_RESULTS, query, searchbox);
+  }
+
+  /**
+   * Dispatches an `sfx::sayt_products_request` event with the provided data.
```
geoffcoutts:
```suggestion
   * Dispatches an [[PRODUCTS_REQUEST_EVENT]] event with the provided data.
```

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -274,7 +426,7 @@ describe('Sayt Component', () => {
     let hideSayt;
 
     beforeEach(() => {
-      dispatchEvent = stub(window, 'dispatchEvent');
+      dispatchEvent = stub(sayt, 'dispatchEvent');
```
geoffcoutts: This whole test section can be removed. It's out of date and duplicated by [the describe block above](https://github.com/groupby/sfx-view/pull/31/files#diff-8340fbc81f5e03de0ad1e53a9141aa2eR240)

packages/web-components/@sfx/products/test/unit/products.test.ts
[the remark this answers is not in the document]
DanielMoniz-GBI: Yep, okay. Makes sense to me.

packages/web-components/@sfx/sayt/src/sayt.ts
jmbtutor: I wasn't going to make any comments about event documentation because we can handle it in the events ticket, but sure ("an" should be changed to "a" if this change is made).

packages/web-components/@sfx/autocomplete/test/unit/autocomplete.test.ts
DanielMoniz-GBI: It's being used by the function under test, even if it's not being used explicitly here. The idea is to not be accidentally adding event listeners, which is what will happen if I remove the stub.

packages/web-components/@sfx/autocomplete/test/unit/autocomplete.test.ts
DanielMoniz-GBI: Same as above. The function under test is using it, so I'm stubbing it out here.

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
DanielMoniz-GBI: @geoffcoutts Excellent catch.

packages/web-components/@sfx/autocomplete/test/unit/autocomplete.test.ts
DanielMoniz-GBI: YOU'RE RIGHT

packages/web-components/@sfx/autocomplete/test/unit/autocomplete.test.ts
DanielMoniz-GBI: Same as above. *sigh*

packages/web-components/@sfx/sayt/src/sayt.ts
geoffcoutts: I only mentioned it since the wrapper function for the autocomplete event handled it. Both were created in this ticket.

PR #33, opened 2019-09-12
SFX-337: Sayt calls are debounced

ssbelle:
Debounces terms and product requests in sayt, as well as autocomplete hover terms so that if several keys are pressed in succession there will not be a query made for each keypress. Instead queries are only made on the leading edge and after the debounceTime has elapsed.  
To test open with demo.html and observe the queries in network tab. 

- Added debounce property to sayt
- Wrapped the `requestSaytAutocompleteTerms` and `requestSaytProducts` in the Debounce method so they will be debounced
- Added tests


packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -77,6 +84,9 @@ export default class Sayt extends LitElement {
     window.addEventListener('click', this.processClick);
     window.addEventListener('keydown', this.processKeyEvent);
     this.setSearchboxListener(this.searchbox, 'add');
+
+    this.requestSaytAutocompleteTerms  = this.debounceSaytRequests(this.requestSaytAutocompleteTerms);
+    this.requestSaytProducts  = this.debounceSaytRequests(this.requestSaytProducts);
```
DanielMoniz-GBI: I'm guessing these lines can happen in the constructor, unless I'm missing a reason why you're waiting for `connectedCallback()`.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -187,6 +197,20 @@ export default class Sayt extends LitElement {
     this.requestSaytProducts(query, searchbox);
   }
 
+  /**
+   * Debounces the function that is passed in the callback paramater.
+   * Customizable delay based on debounceTime attribute on Sayt element.
+   *
+   * @param callback The function to debounce.
+   */
+  debounceSaytRequests(callback: (query: string, searchbox?: string) => void) {
```
DanielMoniz-GBI: My feeling here is that `debounceSaytRequests` doesn't care about the type of the callback.

Either way, we've been leaving out void return type, and just letting that be implicit if there is no provided return type.

I'm wondering if this method should just be called `getDebounce()` or something. Although we use it for debouncing SAYT requests, it doesn't really matter what uses it. I suppose we do use `this.debounceTime`, but even that is not specific to SAYT requests.

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -52,6 +53,18 @@ describe('Sayt Component', () => {
       expect(setSearchboxListener).to.be.calledWith(searchbox, 'add');
       expect(addEventListener).to.be.calledWith(HOVER_AUTOCOMPLETE_TERM_EVENT, sayt.handleAutocompleteTermHover);
     });
+
+    it('should call `debounceSaytRequests` for each of the two debounced methods', () => {
```
DanielMoniz-GBI: We haven't been using backticks in our test descriptions as far as I know (I've also added () below):
```suggestion
    it('should call debounceSaytRequests() for each of the two debounced methods', () => {
```

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -267,6 +280,36 @@ describe('Sayt Component', () => {
     });
   });
 
+  describe('debounceSaytRequests', () => {
```
DanielMoniz-GBI: Most of our descriptions for tests against methods have been using the `()` after them. So:
```suggestion
  describe('debounceSaytRequests()', () => {
```

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -267,6 +280,30 @@ describe('Sayt Component', () => {
...
+      const requestSaytAutocompleteTerms = debounceStub(sayt.requestSaytAutocompleteTerms(), delay, debounceSettings);
+      const requestSaytProducts = debounceStub(sayt.requestSaytProducts(), delay, debounceSettings);
+      const autocompleteTermsReturn = debounceSaytRequests.returns(requestSaytAutocompleteTerms);
+      const productsReturn = debounceSaytRequests.returns(requestSaytProducts);
+
+      sayt.debounceSaytRequests(sayt.requestSaytAutocompleteTerms);
+      sayt.debounceSaytRequests(sayt.requestSaytProducts);
+
+      expect(debounceStub).to.be.calledTwice;
+      expect(debounceSaytRequests).to.be.calledTwice;
```
DanielMoniz-GBI: We've been avoiding asserting exactly how many times a method has been called, and have instead been relying on asserting (as you have done below) that the right arguments are passed to methods.

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -267,6 +280,30 @@ describe('Sayt Component', () => {
...
+      expect(debounceStub).to.be.calledWithExactly(requestSaytAutocompleteTerms, delay, debounceSettings);
+      expect(debounceStub).to.be.calledWithExactly(requestSaytProducts, delay, debounceSettings);
+      expect(debounceSaytRequests).to.equal(autocompleteTermsReturn);
+      expect(debounceSaytRequests).to.equal(productsReturn);
```
DanielMoniz-GBI: I don't believe we're asserting anything of value on the above two lines. I'm curious what your thought process is, but as far as I can tell, you're actually defining these things to be equal because called `.returns()` on a stub returns the stub itself. So you define `debounceSaytRequests` to be a stub, and then later make a new variable equal to `debounceSaytRequests.returns(...)`, so that new variable is definitionally identical to `debounceSaytRequests`.

packages/web-components/@sfx/sayt/package.json
```diff
@@ -20,6 +20,7 @@
   "author": "",
   "license": "MIT",
   "devDependencies": {
-    "html-entities": "^1.2.1"
+    "html-entities": "^1.2.1",
+    "lodash.debounce": "^4.0.8"
```
ssbelle: @jmbtutor I'd like to consult you on using lodash.debounce vs just lodash. My instinct says that its best to use the smaller specific package, however I had some trouble importing it and had to use require, but then could not get it to work for the tests. I could use your input on this, Thanks!

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -1,5 +1,6 @@
 import { LitElement, customElement, html, property, PropertyValues } from 'lit-element';
 import { PRODUCTS_RESPONSE_EVENT, PRODUCTS_REQUEST_EVENT } from '@sfx/products';
+import * as Lodash from 'lodash';
```
jmbtutor: This line is going to go away if you take my suggestion about not using Lodash, so no action needed here, but for the record, `Lodash` is conventionally `_`. As well, you should use the `lodash.debounce` import that you specified earlier instead  of the whole `lodash` package.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -187,6 +197,20 @@ export default class Sayt extends LitElement {
     this.requestSaytProducts(query, searchbox);
   }
 
+  /**
+   * Debounces the function that is passed in the callback paramater.
+   * Customizable delay based on debounceTime attribute on Sayt element.
+   *
+   * @param callback The function to debounce.
+   */
+  getDebounce(callback: any) {
```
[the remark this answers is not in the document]
ssbelle: @jmbtutor it did complain, it threw errors where we are using the method ex: this.getDebounce(this.requestSaytAutocompleteTerms) said it wasn't assignable to the types of the method requestSaytAutocompleteTerms I think . `Type 'Function & { clear(): void; } & { flush(): void; }' is not assignable to type '(query: string, searchbox?: string) => void'.` 

Also is there currently a utils folder? I thought there was but could not see one

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -182,11 +195,33 @@ export default class Sayt extends LitElement {
...
+      false
+    );
+  }
+
+  /**
+   * Dispatches getDebounce for each of the methods to be debounced.
+   * Is triggered once in constructor and when the debounce attribute is changed.
+   */
+  setDebounce() {
+    this.requestSaytAutocompleteTerms  = this.getDebounce(this.requestSaytAutocompleteTerms);
```
geoffcoutts: What happens to the debounce when changing the wait time after it has already been set? The assumption is that `this.requestSaytAutocompleteTerms` will hold reference to a debounced method when being passed to `getDebounce()` a second time. 

Ex. First debounce being 300ms, second debounce setting is 400ms, resulting in a 700ms debounce. 

If this acts like so, a possible solution would be to name the original `requestSaytAutocompleteTerms()` and `requestSaytProducts` in a pseudo private method (ie. `_requestSaytAutocompleteTerms()`) and then continuing to assign the debounced method to the same property.

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -22,6 +23,17 @@ describe('Sayt Component', () => {
       expect(sayt).to.be.an.instanceof(LitElement);
     });
 
+    it('should debounce certain methods', () => {
```
geoffcoutts: This test is currently testing the internals of `setDebounce()`. The test should only assert that it was called (what the `constructor` does), not what happens inside of it.

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -267,6 +279,22 @@ describe('Sayt Component', () => {
...
 
+  describe('getDebounce()', () => {
+    it('should return a debounced callback method', () => {
+      const delay = sayt.debounce;
+      const debounceSettings = false;
+      const termsCallback = stub(sayt, 'requestSaytAutocompleteTerms');
+      const debounceStub = stub(Debounce, 'debounce');
+      const expectedDebouncedFunction = () => 123;
+
+      debounceStub.returns(expectedDebouncedFunction);
```
geoffcoutts: This should be part of the arrange (first) section of the test. The stub is created in the test so there's no need to change it later.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -114,6 +124,9 @@ export default class Sayt extends LitElement {
       this.setSearchboxListener(oldSearchbox, 'remove');
       this.setSearchboxListener(this.searchbox, 'add');
     }
+    if (changedProps.has('debounce')) {
```
geoffcoutts: Should update the tests for `updated()` to account for this.

packages/web-components/@sfx/sayt/src/sayt.ts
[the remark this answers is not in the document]
jmbtutor: Looks like we need a generic. Looking at the [type signature for `debounce`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c9912af2baf0cad74fd61fd824bc653ee3c721a2/types/debounce/index.d.ts#L12), we might have to use a generic:

```suggestion
  getDebounce<A extends Function>(callback: A): ReturnType<typeof debounce<A>> {
```

That way, the return value, which is of type `(query: string, searchbox?: string) => void & { clear(): void; } & { flush(): void; }` is compatible with the original type `(query: string, searchbox?: string) => void`.

Note that I haven't tested `ReturnType<typeof debounce<A>>`, so you might have to play around with it before it works. If it doesn't work with `typeof` try removing it. If it still doesn't work, let me know and we can work it out together.

> Also is there currently a utils folder?

If there isn't, you can make one.

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -22,6 +23,14 @@ describe('Sayt Component', () => {
       expect(sayt).to.be.an.instanceof(LitElement);
     });
 
+    it('should call setDebounce()', () => {
+      const setDebounceStub = stub(Sayt.prototype, 'setDebounce');
```
geoffcoutts: As a general comment for changes throughout the file, we have gone with the pattern of not including "Stub" in the name of stubbed functions.

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -267,6 +287,33 @@ describe('Sayt Component', () => {
     });
   });
 
+  describe('getDebounce()', () => {
+    it('should return a debounced callback method', () => {
+      const delay = sayt.debounce;
```
jmbtutor: I think it would be better to set this to a known concrete value. For example, you can set it to 200.

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -267,6 +287,33 @@ describe('Sayt Component', () => {
     });
   });
 
+  describe('getDebounce()', () => {
+    it('should return a debounced callback method', () => {
```
jmbtutor:
```suggestion
    it('should return a debounced callback method with the configured delay', () => {
```

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -350,3 +393,17 @@ export default class Sayt extends LitElement {
     `;
   }
 }
+
+/**
+ * The type of the callback expected to be passed to getDebounce.
+ */
+export interface requestSaytType extends debounce {
```
jmbtutor: In general, interfaces should start with a capital letter. That said, I don't think we'll need this interface in the new approach.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -42,6 +43,19 @@ export default class Sayt extends LitElement {
...
    */
   @property({ type: Number, reflect: true }) minSearchLength = 3;
+  /**
+   * The debounce delay in millilseconds.
+   */
+  @property({ type: Number, reflect: true }) debounce = 300;
+  /**
+   * The debouncedRequestSaytProducts method that will wrap the requestSaytProducts method in a debounce.
+   */
+  debouncedRequestSaytProducts: requestSaytType;
```
jmbtutor: These declarations won't be necessary with the new approach, but you'll have to cast the functions as `ReturnType<typeof debounce>` in the tests to be able to use `flush()`.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -61,6 +75,10 @@ export default class Sayt extends LitElement {
     this.processSfxSearchboxChange = this.processSfxSearchboxChange.bind(this);
     this.setSearchboxListener = this.setSearchboxListener.bind(this);
     this.handleAutocompleteTermHover = this.handleAutocompleteTermHover.bind(this);
+    this.getDebounce = this.getDebounce.bind(this);
+    this.setDebounce = this.setDebounce.bind(this);
```
jmbtutor: `get` and `set` aren't that clear to me for this purpose; in particular, `setDebounce` suggests to me that you're setting the `debounce` property, which isn't what it's doing. `setDebounce` can be called `setDebouncedMethods` or something like that, and I suggest later on in this review that we probably don't need `getDebounce`.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -182,9 +203,31 @@ export default class Sayt extends LitElement {
...
 
-    this.requestSaytAutocompleteTerms(query, searchbox);
-    this.requestSaytProducts(query, searchbox);
+  /**
+   * Debounces the function that is passed in the callback paramater.
+   * Customizable delay based on debounceTime attribute on Sayt element.
+   *
+   * @param callback The function to debounce.
+   */
+  getDebounce<requestSaytType extends Function>(callback: requestSaytType): ReturnType<debounce> {
```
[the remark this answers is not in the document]
jmbtutor: But actually, I don't think we even need this. Since we have `setDebounce` now, we can just use `debounce` directly and remove this function.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -182,9 +203,31 @@ export default class Sayt extends LitElement {
...
+    );
+  }
+
+  /**
+   * Dispatches getDebounce for each of the methods to be debounced.
+   * Is triggered once in constructor and when the debounce attribute is changed.
+   */
+  setDebounce() {
+    this.debouncedRequestSaytAutocompleteTerms  = this.getDebounce(this.requestSaytAutocompleteTerms);
+    this.debouncedRequestSaytProducts  = this.getDebounce(this.requestSaytProducts);
```
jmbtutor: Since `getDebounce` is only being used in this function, I don't think we need to make it a separate method. We can use `debounce` directly:

```js
    this.debouncedRequestSaytAutocompleteTerms = debounce(this.requestSaytAutocompleteTerms, this.debounce, false);
    this.debouncedRequestSaytProducts = debounce(this.requestSaytProducts, this.debounce, false);
```

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -182,9 +203,31 @@ export default class Sayt extends LitElement {
       this.hideSayt();
       return;
     }
+    this.debouncedRequestSaytAutocompleteTerms(query, searchbox);
+    this.debouncedRequestSaytProducts(query, searchbox);
```
jmbtutor: The `handleAutocompleteTermHover` function should also be using `debouncedRequestSaytProducts`.

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -47,6 +48,19 @@ export default class Sayt extends LitElement {
...
    */
   @property({ type: Number, reflect: true }) minSearchLength = 3;
+  /**
+   * The debounce delay in millilseconds.
+   */
+  @property({ type: Number, reflect: true }) debounce = 300;
+  /**
+   * The debouncedRequestSaytProducts method that will wrap the requestSaytProducts method in a debounce.
+   */
+  debouncedRequestSaytProducts: RequestSaytType;
```
ssbelle: @jmbtutor You mentioned in one of your reviews that this shouldn't be necessary, do you have any further direction as to how to set these debounced methods?

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -47,6 +48,19 @@ export default class Sayt extends LitElement {
...
+  /**
+   * The debouncedRequestSaytAutocompleteTerms method that will wrap the requestSaytAutocompleteTerms method in a debounce.
+   */
+  debouncedRequestSaytAutocompleteTerms: RequestSaytType;
```
ssbelle: Same as comment above

packages/web-components/@sfx/sayt/src/sayt.ts
ssbelle: It does not below on line 266 if that is correct

packages/web-components/@sfx/sayt/package.json
```diff
@@ -19,9 +19,11 @@
   "author": "",
   "license": "MIT",
   "devDependencies": {
+    "@types/debounce": "^1.2.0",
```
jmbtutor: I don't remember if I recommended that this should be here, but since we'll exporting a type that requires the type of `debounce` (if you implement my suggested type), this has to be listed under `dependencies`. This [StackOverflow post](https://stackoverflow.com/a/46011417) has more detail.

packages/web-components/@sfx/sayt/test/unit/sayt.test.ts
```diff
@@ -269,16 +289,31 @@ describe('Sayt Component', () => {
...
+    it('should set debounced methods for requestSaytAutocompleteTerms and requestSaytProducts', () => {
+       const debounce = stub(Debounce, 'debounce');
+       const expectedDebouncedFunction = () => 123;
+       debounce.returns(expectedDebouncedFunction);
+
+       sayt.setDebouncedMethods();
+
+       expect(debounce).to.be.calledTwice;
+       expect(sayt.debouncedRequestSaytAutocompleteTerms).to.equal(expectedDebouncedFunction);
+       expect(sayt.debouncedRequestSaytProducts).to.equal(expectedDebouncedFunction);
```
jmbtutor: This test isn't quite robust. The test needs to assert that `debounce` was called with the right arguments, and that the correct invocations assign to the correct properties.

You can use sinon's `withArgs` to do that:

```typescript
const debounce = stub(Debounce, 'debounce');
const delay = sayt.debounce = 200;
const debouncedRequestSaytAutocompleteTerms = () => 123;
const debouncedRequestSaytProducts = () => 456;
debounce.withArgs(sayt.requestSaytAutocompleteTerms, delay, false).returns(debouncedRequestSaytAutocompleteTerms);
debounce.withArgs(sayt.requestSaytProducts, delay, false).returns(debouncedRequestSaytProducts);

sayt.setDebouncedMethods();

expect(sayt.debouncedRequestSaytAutocompleteTerms).to.equal(debouncedRequestSaytAutocompleteTerms);
expect(sayt.debouncedRequestSaytProducts).to.equal(debouncedRequestSaytProducts);
```

The stub will only return the correct value if it's called with the correct arguments.

PR #36, opened 2019-09-23
SFX-385: Allow for passing custom product transformer

DanielMoniz-GBI:
This PR primarily updates the demo to define functions for transforming the product for Search and Sayt, much as the client would have to do. The transformer for Apparel, previously hardcoded in the Logic Layer, is now defined in `demo.html`.

Note that the product transformer for Search is intentionally simple. It will display only product titles.

Additionally:
* Update Product and Searchbox components to extend `LitElement` instead of `Base`. This removes issues with multiple renders/updates of components (artifacts from previous renders remain in the DOM).
* Add event listeners for events in the demo file. This is intentional. It is easier to demonstrate functionality, and to debug issues, if we can see events being logged to the console.

### DIFF
--- a/demo/demo.html
+++ b/demo/demo.html
@@ -7,8 +7,8 @@
     <script src="./temp/sfx-all-plugins-bundle.js"></script>
   </head>
   <body>
-    <sfx-search-box placeholder="search" clearbutton searchbutton></sfx-search-box>
-    <sfx-sayt visible></sfx-sayt>
+    <sfx-search-box id="main-search" placeholder="search" clearbutton searchbutton></sfx-search-box>
+    <sfx-sayt searchbox="main-search" visible></sfx-sayt>
     <sfx-list class="product-grid"></sfx-list>
 
     <script>
@@ -67,6 +67,10 @@
       });
 
       window.addEventListener('sfx::sayt_products_response', (event) => {
+        console.log(event.type, event.detail);
+      });
+
+      window.addEventListener('sfx::sayt_products_error', (event) => {
         console.log(event.type, event.detail);
       });
     </script>
--- a/demo/demo.html
+++ b/demo/demo.html
@@ -46,7 +46,7 @@
       }
 
       function searchProductTransformer(record) {
-        return { title: record.title };
+        return { label: record.title };
       }
 
       // init core and plugins
@@ -58,11 +58,12 @@
       const searchDriverPlugin = new sfxPlugins.SearchDriverPlugin({ productTransformer: searchProductTransformer });
       // register all plugins with core
       core.register([saytPlugin, saytDriverPlugin, domEventsPlugin, searchPlugin, searchDriverPlugin]);
+
       // search integration/functionality test
       window.addEventListener('sfx::search_response', (e) => {
         const productsGrid = document.querySelector('sfx-products');
         if (productsGrid) {
-          productsGrid.products = e.detail.products;
+          productsGrid.items = e.detail.products;
         }
       });
 
--- a/demo/demo.html
+++ b/demo/demo.html
@@ -57,9 +57,9 @@
       // init core and plugins
       const core = new sfxCore.Core();
       const domEventsPlugin = new sfxPlugins.DomEventsPlugin()
-      const saytPlugin = new sfxPlugins.SaytPlugin({subdomain: 'apparel', collection: 'products'});
+      const saytPlugin = new sfxPlugins.SaytPlugin({ subdomain: 'apparel' });
       const saytDriverPlugin = new sfxPlugins.SaytDriverPlugin({ productTransformer });
-      const searchPlugin = new sfxPlugins.SearchPlugin({ customerId: 'apparel', collection: 'products', area: 'Storefront'});
+      const searchPlugin = new sfxPlugins.SearchPlugin({ customerId: 'apparel' });
       const searchDriverPlugin = new sfxPlugins.SearchDriverPlugin({ productTransformer: searchProductTransformer });
       // register all plugins with core
       core.register([saytPlugin, saytDriverPlugin, domEventsPlugin, searchPlugin, searchDriverPlugin]);
--- a/demo/demo.html
+++ b/demo/demo.html
@@ -46,7 +46,7 @@
       }
 
       function searchProductTransformer(record) {
-        return { title: 'product 1' };
+        return { title: record.title + '!!!' };
       }
 
       // init core and plugins
--- a/demo/demo.html
+++ b/demo/demo.html
@@ -9,10 +9,10 @@
 
   <body>
     <sfx-search-box id="main-search" placeholder="search" clearbutton searchbutton></sfx-search-box>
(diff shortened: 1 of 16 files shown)


demo/demo.html
```diff
@@ -4,29 +4,72 @@
     <title>Demo of Logic and View Integration</title>
     <script src="../dist/components.js"></script>
     <script src="https://cdn.groupbycloud.com/sfx-all-plugins-bundle.js"></script>
+    <!-- <script src="./temp/sfx-all-plugins-bundle.js"></script> -->
```
DanielMoniz-GBI: We need this often enough that I think it makes sense to leave in.

demo/demo.html
```diff
@@ -4,29 +4,72 @@
...
+
   <body>
-    <sfx-search-box placeholder="search" clearbutton searchbutton></sfx-search-box>
-    <sfx-sayt visible></sfx-sayt>
-    <sfx-list class="product-grid"></sfx-list>
+    <sfx-search-box id="main-search" placeholder="search" clearbutton searchbutton></sfx-search-box>
+    <sfx-sayt searchbox="main-search" visible></sfx-sayt>
+
+    <sfx-search-box id="secondary-search" group="secondary" placeholder="search" clearbutton searchbutton></sfx-search-box>
+    <sfx-sayt searchbox="secondary-search" group="secondary" visible></sfx-sayt>
```
DanielMoniz-GBI: This is a second searchbox and Sayt for demoing purposes. I can take it out if there is a reason to do so, but we often want to demo against two components. Additionally, end-to-end testing will likely require two components in order to prove that (say) multiple `<sfx-sayt>`s are able to act indepedently.

demo/demo.html
```diff
@@ -4,29 +4,72 @@
...
+
+        return {
+          title: data.title,
+          price: nonvisualVariants[0].originalPrice,
+          imageSrc: firstVariant.productImage,
+          imageAlt: data.title,
+          productUrl: firstVariant.productImage,
+          // @TODO Handle variants
+        }
+      }
```
DanielMoniz-GBI: This code is almost line-for-line from the old SaytDriver.

demo/demo.html
```diff
@@ -4,29 +4,72 @@
...
       const saytPlugin = new sfxPlugins.SaytPlugin({subdomain: 'apparel', collection: 'products'});
-      const saytDriverPlugin = new sfxPlugins.SaytDriverPlugin();
+      const saytDriverPlugin = new sfxPlugins.SaytDriverPlugin({ productTransformer });
       const searchPlugin = new sfxPlugins.SearchPlugin({ customerId: 'apparel', collection: 'products', area: 'Storefront'});
-      const searchDriverPlugin = new sfxPlugins.SearchDriverPlugin();
+      const searchDriverPlugin = new sfxPlugins.SearchDriverPlugin({ productTransformer: searchProductTransformer });
       // register all plugins with core
       core.register([saytPlugin, saytDriverPlugin, domEventsPlugin, searchPlugin, searchDriverPlugin]);
-      // search integration/functionality test
-      window.addEventListener('sfx::search_response', (e) => {
```
DanielMoniz-GBI: test

demo/demo.html
```diff
@@ -4,29 +4,72 @@
...
-        document.getElementsByClassName('product-grid')[0]
-        .setAttribute('items', JSON.stringify(testListResults));
+
+      window.addEventListener('sfx::sayt_products_response', (event) => {
+        console.log(event.type, event.detail);
+      });
+
+      window.addEventListener('sfx::sayt_products_error', (event) => {
+        console.log(event.type, event.detail);
       });
```
DanielMoniz-GBI: I can remove these on request, but it seems reasonable (and time saving) to have a bit of logging in our demo.

DanielMoniz-GBI:
I updated `demo/temp/sfx-all-plugins-bundle.js` because it seemed odd purposely leaving an out-of-date bundle. We should either keep this up to date or remove it entirely. I'm open to suggestions!

jmbtutor:
The build fails. I think you need to update the version of `@sfx/events` in the appropriate `package.json` files so you pull in the new generic. It also needs to be rebased.

Between updating the Logic bundle regularly and removing it, I'd prefer to remove it so that we don't unnecessarily bloat the size of our repo, but we can keep the commented script tag in. I'm okay with updating it for this PR, though.

demo/demo.html
jmbtutor: I think this is a good thing to have. We can add a bit of styling later to make it look somewhat saner (so that it doesn't shift past the end of the page, for example).

demo/demo.html
jmbtutor: We should consider having listeners for all our events. Maybe later.

Plugin idea: we have a plugin that depends on a Logger plugin (yet to be built) and the Events plugin. It would subscribe to a list of events and log out its name and payload upon receipt.

demo/demo.html
```diff
@@ -4,29 +4,72 @@
...
-    <sfx-search-box placeholder="search" clearbutton searchbutton></sfx-search-box>
-    <sfx-sayt visible></sfx-sayt>
-    <sfx-list class="product-grid"></sfx-list>
+    <sfx-search-box id="main-search" placeholder="search" clearbutton searchbutton></sfx-search-box>
+    <sfx-sayt searchbox="main-search" visible></sfx-sayt>
+
+    <sfx-search-box id="secondary-search" group="secondary" placeholder="search" clearbutton searchbutton></sfx-search-box>
+    <sfx-sayt searchbox="secondary-search" group="secondary" visible></sfx-sayt>
+
+    <sfx-products></sfx-products>
```
geoffcoutts: Since you have duplicated the searchbox and sayt, is it worth duplicating the products grid as well? Not required at all.

demo/demo.html
```diff
@@ -4,29 +4,72 @@
...
     <script src="https://cdn.groupbycloud.com/sfx-all-plugins-bundle.js"></script>
+    <!-- <script src="./temp/sfx-all-plugins-bundle.js"></script> -->
   </head>
+
   <body>
-    <sfx-search-box placeholder="search" clearbutton searchbutton></sfx-search-box>
-    <sfx-sayt visible></sfx-sayt>
-    <sfx-list class="product-grid"></sfx-list>
+    <sfx-search-box id="main-search" placeholder="search" clearbutton searchbutton></sfx-search-box>
+    <sfx-sayt searchbox="main-search" visible></sfx-sayt>
```
geoffcoutts: The `visible` attribute should probably be dropped from these as the default.

demo/demo.html
DanielMoniz-GBI: As it is, one of the searches populates the grid and the other does not. That's valuable to me. This is a reasonable idea, though.

demo/demo.html
DanielMoniz-GBI: I do like this idea. We could instantiate it on our demo and show clients that they can always use it to understand how SF-X is hooking up.

PR #37, opened 2019-09-30
SFX-379 - Emit all data from SAYT responses

ChantalDesrochers:
- update manufactured getSaytProductsResponseEvent in Storybook to have type 'any'
- update @sfx/event repo reference to most recent commit

### DIFF
--- a/.storybook/common.ts
+++ b/.storybook/common.ts
@@ -140,6 +140,36 @@ export function getSaytProductsResponseEvent(): CustomEvent<SaytProductsResponse
   return new CustomEvent(SAYT_PRODUCTS_RESPONSE, {
     detail: {
       products: getProducts(5),
+      originalResponse: {
+        area: "Production",
+        availableNavigation: [{name: "categories.3", displayName: "Sub Category", type: "Value", refinements: [{type: "Value", count: 616, value: "Women"}], metaData: []}],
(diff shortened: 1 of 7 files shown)


.storybook/common.ts
```diff
@@ -117,10 +116,40 @@ export const saytProductsResponseEvent = {
   }
 };
 
-export function getSaytProductsResponseEvent(): CustomEvent<SaytProductsResponsePayload> {
+export function getSaytProductsResponseEvent(): CustomEvent<any> {
```
jmbtutor: Ah, that's what you meant by `any`.  You should scope your `any` as small as possible, so you should keep this as-is and cast the object given to `originalResponse` as `any`.

.storybook/common.ts
```diff
@@ -117,10 +116,40 @@ export const saytProductsResponseEvent = {
...
+        records: [],
+        redirect: "",
+        relatedQueries: [],
+        rewrites: [],
+        selectedNavigation: [],
+        siteParams: [],
+        template: {ruleName: "none", name: "default", zones: {}},
+        totalRecordCount: 2451,
+        warnings: "",
+      }
```
jmbtutor: In case the suggestion above wasn't clear, this is the second half of the suggestion:

```suggestion
      } as any,
```

PR #39, opened 2019-10-16
SFX-381: Add product variants to product transformer

ChantalDesrochers:
- add parseVariants function to demo.html (called by productTransformer which is passed into sayt-driver
- removed Base extension and slot reference from product and variant components
- updated tests and documentation accordingly
- NOTE: ran formatter on demo file

### DIFF
--- a/demo/demo.html
+++ b/demo/demo.html
@@ -5,6 +5,27 @@
     <script src="../dist/components.js"></script>
     <!-- <script src="https://cdn.groupbycloud.com/sfx-all-plugins-bundle.js"></script> -->
     <script src="./temp/sfx-all-plugins-bundle.js"></script>
+    <!-- <link href="../themes/sfx-elegant-them/components/_product-variant.scss" rel="stylesheet" /> -->
+    <style>
+      sfx-product-variant {
+        width: 30px;
+        height: 30px;
+        display: inline-block;
+        background-size: cover;
+        margin: 2px;
+        overflow: hidden;
+        cursor: pointer;
+      }
+
+      sfx-product-variant img {
+        width: auto;
+        height: auto;
+        min-width: 30px;
+        min-height: 30px;
+        max-width: 100%;
+        max-height: 100%;
+      }
+    </style>
   </head>
 
   <body>
@@ -24,7 +45,7 @@
 
     <script>
       function parseRecord(record) {
-        console.log('record', record)
+        console.log("record", record);
         const data = record.allMeta;
         const firstVariant = data.visualVariants[0];
         const nonvisualVariants = firstVariant.nonvisualVariants;
@@ -38,13 +59,12 @@
           visualVariants
         };
       }
-      // questions - should we be using swatch colors? just checked the events repo and it says color should be a valid CSS color...
-      // image? that's different than product image? no thumbnail image in the records?
-      // they would be determining whether or not they want to use text, color or image as the type?
+
       function parseVariants(variants) {
         return variants.map(v => ({
-          color: v.swatchImage || v.standardColor,
+          color: v.standardColor,
           text: v.standardColor,
+          image: v.swatchImage,
           product: {
             imageSrc: v.productImage,
             imageAlt: v.description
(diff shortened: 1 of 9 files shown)


packages/web-components/@sfx/product/src/product.ts
```diff
@@ -108,13 +105,8 @@ export default class Product extends LitElement {
     : ''
 }
         </ul>
-      </slot>
-      <slot name="title">
         ${this.urlWrap(productUrl, html`<h3 class="sfx-title">${title}</h3>`)}
-      </slot>
-      <slot name="price">
         <p class="sfx-price">${price}</p>
-      </slot>
```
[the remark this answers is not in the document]
ChantalDesrochers: I've updated the themes, replacing references to slots.

demo/demo.html
```diff
@@ -21,33 +47,51 @@
...
+        const { data, firstVariant, nonvisualVariants, visualVariants } = filter;
         return {
           title: data.title,
           price: nonvisualVariants[0].originalPrice,
           imageSrc: firstVariant.productImage,
           imageAlt: data.title,
           productUrl: firstVariant.productImage,
-          // @TODO Handle variants
-        }
+          ...(visualVariants.length > 1 && {
```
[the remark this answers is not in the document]
ChantalDesrochers: If visulaVariants is less than 1 (e.g. there are no visualVariants), the variants property is not added to the object. Is this not what we want?

packages/web-components/@sfx/product/src/variant.ts
```diff
@@ -44,16 +43,19 @@ export default class Variant extends Base {
   render(): TemplateResult {
     const { variant } = this;
 
-    return html`
-      ${this.type === 'image'
-    ? html`<img
+    return html`${this.type === 'image'
```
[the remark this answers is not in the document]
ChantalDesrochers: I made an update - is this what you meant?

packages/web-components/@sfx/product/src/variant.ts
```diff
@@ -42,15 +42,13 @@ export default class Variant extends LitElement {
 
   render(): TemplateResult {
     const { variant } = this;
+    const img = this.type === 'image' ? html`<img
+    class="sfx-variant-image"
+    src="${variant.image}"
+    alt="${variant.text}"
+  />` : '';
```
jmbtutor: Yes, this is what I meant, but I think we can improve the formatting. Maybe something like this?
```js
    const img = this.type === 'image'
      ? html`<img
          class="sfx-variant-image"
          src="${variant.image}"
          alt="${variant.text}"
        />`
      : '';
```

demo/demo.html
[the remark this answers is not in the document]
jmbtutor: I should double-check my facts first next time. It looks like [`false` does spread to nothing](https://github.com/tc39/proposal-object-rest-spread/issues/45). This is fine as-is.

packages/web-components/@sfx/product/src/variant.ts
```diff
@@ -43,17 +42,20 @@ export default class Variant extends Base {
...
           class="sfx-variant-image"
           src="${variant.image}"
           alt="${variant.text}"
         />`
-    : ''
-}
+      : '';
+
+    return html`${img}
       ${this.type === 'text' ? variant.text : ''}
```
geoffcoutts: Is it worth formatting this the same way as `img` for consistency? If text was a little more complex I would say yes for sure, but as it is, i'm not too concerned either way

jmbtutor: If we want to be really clean, yes, but I'm not too concerned about it as long as it doesn't span multiple lines and we don't need to reuse it.

demo/demo.html
```diff
@@ -5,31 +5,71 @@
...
 
         return {
           data,
           firstVariant,
           nonvisualVariants,
+          visualVariants,
         };
       }
 
+      function parseVariants(variants) {
```
ssbelle: @ChantalDesrochers  @jmbtutor  Just thinking, do we want to be building out so much js and css in the html file? I understand that this is meant to approximate customizations done by the client but would it be cleaner to abstract it to a 'transformer.js' file and import that?

jmbtutor: I'm not too concerned about it because this is only for our use, it's only used here, and we're probably not going to change it that often. In my opinion, the demo file should be clean enough that we can work in it, but it doesn't have to be so clean that it's exemplary.

PR #40, opened 2019-10-18
SFX-380: Enable using arrow keys to change autocomplete selection

jmbtutor:
While in the searchbox with an open sayt, pressing the Up and Down arrow keys will change the selected autocomplete term and accordingly update the products shown.

### DIFF
--- a/packages/web-components/@sfx/autocomplete/README.md
+++ b/packages/web-components/@sfx/autocomplete/README.md
@@ -4,15 +4,28 @@
 
 The Autocomplete component displays search terms related to a query term. It also dispatches events whenever these search terms are interacted with.
 
-This component listens for and dispatches a number of events. These events are defined in the [`@sfx/events`][sfx-events] package.
+### Term Selection
+
+At most one term is selected at any given time. When a term is selected,
+the `aria-selected` property for the item is set to
+
+The selection can be changed in one of three ways:
+
+* Hovering over a term
+* Using `selectNext()` and `selectPrevious()`
+* Setting the `selectedIndex` property
 
 ### Recieved Events
+
+This component listens for a number of events. These events are defined in the [`@sfx/events`][sfx-events] package.
 
 #### `AUTOCOMPLETE_RESPONSE`
 
 Upon receiving this event, the `sfx-autocomplete` component will populate its `results` property with search terms.
 
 ### Dispatched Events
+
+This component dispatches a number of events. These events are defined in the [`@sfx/events`][sfx-events] package.
 
 #### `AUTOCOMPLETE_ACTIVE_TERM`
 
--- a/packages/web-components/@sfx/autocomplete/README.md
+++ b/packages/web-components/@sfx/autocomplete/README.md
@@ -6,14 +6,26 @@ The Autocomplete component displays search terms related to a query term. It als
 
 ### Term Selection
 
-At most one term is selected at any given time. When a term is selected,
-the `aria-selected` property for the item is set to
+At most one term is selected at any given time.
+When a term is selected, the `aria-selected` property for the item is set to `true`.
+
+The `selectedIndex` property determines which item is selected;
+this value is the 0-based index of the selected item relative to the first item of the entire results.
+This property can be set to change the selection.
+If the value is outside the range of the results, no item will be selected.
+
+The read-only `selectedId` property contains the DOM ID of the currently selected item.
 
 The selection can be changed in one of three ways:
 
-* Hovering over a term
-* Using `selectNext()` and `selectPrevious()`
-* Setting the `selectedIndex` property
+* Hovering over a term.
+  Hovering over a term will change the selection immediately.
+* Using `selectNext()` and `selectPrevious()`.
+  These methods will select the next and previous items in the list respectively.
+  If the selection falls off one end, it will wrap around to the other end.
+* Setting the `selectedIndex` property.
+  The item at the index given by `selectedIndex` will be selected.
+  If there is no item at that index, no items will be selected.
 
 ### Recieved Events
 
--- a/packages/web-components/@sfx/autocomplete/README.md
+++ b/packages/web-components/@sfx/autocomplete/README.md
@@ -12,7 +12,7 @@ When a term is selected, the `aria-selected` property for the item is set to `tr
 The `selectedIndex` property determines which item is selected;
 this value is the 0-based index of the selected item relative to the first item of the entire results.
 This property can be set to change the selection.
-If the value is outside the range of the results, no item will be selected.
+If the value is outside the range of results (including negative values), no item will be selected.
 
 The read-only `selectedId` property contains the DOM ID of the currently selected item.
 
--- a/packages/web-components/@sfx/autocomplete/README.md
+++ b/packages/web-components/@sfx/autocomplete/README.md
@@ -25,7 +25,8 @@ The selection can be changed in one of three ways:
   If the selection falls off one end, it will wrap around to the other end.
 * Setting the `selectedIndex` property.
   The item at the index given by `selectedIndex` will be selected.
-  If there is no item at that index, no items will be selected.
+  If there is no item at that index (i.e. if `selectedIndex` is negative or if
+  it is past the end of the list), no items will be selected.
 
 ### Recieved Events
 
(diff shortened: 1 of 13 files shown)


packages/web-components/@sfx/autocomplete/src/autocomplete.ts
```diff
@@ -77,14 +138,16 @@ export default class Autocomplete extends LitElement {
...
-   *
-   * @param event An event that contains a Sayt autocomplete term.
+   * Dispatches an [[AUTOCOMPLETE_ACTIVE_TERM]] event with the selected term.
    */
-  handleHoverTerm(event: MouseEvent): void {
-    const target = event.target as HTMLElement;
-    if (target.tagName.toLowerCase() !== 'li') return;
-    const term = target.innerText;
+  dispatchSelectedTerm(): void {
+    if (this.selectedIndex < 0 || this.selectedIndex >= this.length) return;
```
jmbtutor: I would have liked to have written this as `this.selectedIndex < 0 || this.length <= this.selectedIndex` to better emphasize the range, but our `yoda` settings currently don't allow ranges. I may submit an amendment to the linter settings in the near future.

packages/web-components/@sfx/autocomplete/README.md
```diff
@@ -4,16 +4,41 @@
...
+* Using `selectNext()` and `selectPrevious()`.
+  These methods will select the next and previous items in the list respectively.
+  If the selection falls off one end, it will wrap around to the other end.
+* Setting the `selectedIndex` property.
+  The item at the index given by `selectedIndex` will be selected.
+  If there is no item at that index, no items will be selected.
 
 ### Recieved Events
 
+This component listens for a number of events. These events are defined in the [`@sfx/events`][sfx-events] package.
```
ssbelle: We mention that there are a number of events but we only list one below? Should there be any documentation here for the other events?

packages/web-components/@sfx/autocomplete/README.md
```diff
@@ -4,16 +4,41 @@
...
+
 #### `AUTOCOMPLETE_RESPONSE`
 
 Upon receiving this event, the `sfx-autocomplete` component will populate its `results` property with search terms.
 
 ### Dispatched Events
 
+This component dispatches a number of events. These events are defined in the [`@sfx/events`][sfx-events] package.
```
ssbelle: same question for here as above

packages/web-components/@sfx/autocomplete/README.md
jmbtutor: Technically one is "a number", but point taken. Autocomplete only listens to the one event, but the wording is there to be consistent with the other READMEs and so that we don't have to remember to change it if we do end up listening to more events.

packages/web-components/@sfx/autocomplete/test/unit/autocomplete.test.ts
```diff
@@ -144,46 +220,146 @@ describe('Autcomplete Component', () => {
...
+      autocomplete.selectedIndex = 12;
+
+      autocomplete.selectPrevious();
+
+      expect(autocomplete.selectedIndex).to.equal(length - 1);
+    });
+  });
+
+  describe('getSetSelectedIndexCallback()', () => {
+    it('shoudld return a handler that sets the selected index', () => {
```
ChantalDesrochers: typo (should)

packages/web-components/@sfx/autocomplete/src/autocomplete.ts
```diff
@@ -38,21 +42,43 @@ export default class Autocomplete extends LitElement {
...
    */
   connectedCallback(): void {
     super.connectedCallback();
+
     window.addEventListener(AUTOCOMPLETE_RESPONSE, this.receivedResults);
+
+    const role = this.getAttribute('role');
+    const roles = role ? role.split(' ') : [];
+    if (!roles.includes('listbox')) {
+      roles.unshift('listbox');
```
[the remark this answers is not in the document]
jmbtutor: `role` is specified as a space-separated list of roles, where the first one is the primary role and any additional ones are fallbacks. By adding `listbox` first, we're asserting that `listbox` is the primary role for this component. We did originally have `push` here, and perhaps that may be correct if you say that we shouldn't be overriding user-set values, but this component is written with the assumption that it's a listbox.

packages/web-components/@sfx/autocomplete/src/autocomplete.ts
```diff
@@ -63,6 +89,41 @@ export default class Autocomplete extends LitElement {
     window.removeEventListener(AUTOCOMPLETE_RESPONSE, this.receivedResults);
   }
 
+  updated(changedProps: PropertyValues): void {
```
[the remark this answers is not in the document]
jmbtutor: I'd prefer not to because TypeDoc is _supposed_ to inherit the docs of the parent class, but it doesn't seem to be doing so, so I'll add one.

packages/web-components/@sfx/autocomplete/src/autocomplete.ts
```diff
@@ -95,11 +166,80 @@ export default class Autocomplete extends LitElement {
...
+    this.selectedIndex = (normalizedSelectedIndex - 1 + this.itemCount) % this.itemCount;
+  }
+
+  /**
+   * Returns a callback that sets [[selectedIndex]] to the given index.
+   *
+   * @param index The index to set.
+   * @returns A callback that sets [[selectedIndex]].
+   */
+  getSetSelectedIndexCallback(index: number): () => void {
```
ssbelle: @jmbtutor I wonder if this `getSetSelectedIndexCallback` method  could be better named. The name is a bit confusing to me, I get by definition you are doing both but I'm not sure if in that case it should be something like `handleSelectedIndexCallback` or follow some other naming convention? What do you think?

ssbelle: We also have started to use the get and set keywords elsewhere in this file and I know there are benefits to that, is this a method that would be better broken out into a get and a set method?

jmbtutor: We were hoping someone would flag this, because we didn't like the name very much, either. This function was so named because it returns a callback that sets the selected index.

> `handleSelectedIndexCallback`

This doesn't work because we're creating a closure that returns a callback. This function doesn't do any handling itself.

> is this a method that would be better broken out into a get and a set method?

No. Because we're creating a closure with a variable, we need to be able to accept that variable, which means it needs to be a function. For this particular name, there is nothing to set to that name (i.e. one is not expected to do anything like `this.getSetSelectedIndexCallback = /*...*/`), so a setter would not be appropriate. For setting `selectedIndex` inside, a setter would be pointless and irrelevant because it is simply a property that holds a value (that LitElement actually turns it into a setter is beside the point).

`createSelectedIndexSetter` is the best alternative I can think of. Thoughts?

packages/web-components/@sfx/autocomplete/README.md
```diff
@@ -4,16 +4,42 @@
...
-This component listens for and dispatches a number of events. These events are defined in the [`@sfx/events`][sfx-events] package.
+### Term Selection
+
+At most one term is selected at any given time.
+When a term is selected, the `aria-selected` property for the item is set to `true`.
+
+The `selectedIndex` property determines which item is selected;
+this value is the 0-based index of the selected item relative to the first item of the entire results.
+This property can be set to change the selection.
+If the value is outside the range of the results, no item will be selected.
```
ssbelle: sorry did we want to mention that 'outside the results' includes a negative value? @jmbtutor

packages/web-components/@sfx/autocomplete/README.md
```diff
@@ -4,16 +4,41 @@
...
+The selection can be changed in one of three ways:
+
+* Hovering over a term.
+  Hovering over a term will change the selection immediately.
+* Using `selectNext()` and `selectPrevious()`.
+  These methods will select the next and previous items in the list respectively.
+  If the selection falls off one end, it will wrap around to the other end.
+* Setting the `selectedIndex` property.
+  The item at the index given by `selectedIndex` will be selected.
+  If there is no item at that index, no items will be selected.
```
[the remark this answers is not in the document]
jmbtutor: I've added some more words to clarify.

packages/web-components/@sfx/autocomplete/README.md
jmbtutor: I can add that in.

packages/web-components/@sfx/autocomplete/src/autocomplete.ts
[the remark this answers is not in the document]
ChantalDesrochers: I personally think more flexibility is better but I understand the listbox assertion as well.

PR #41, opened 2019-10-18
SFX-426: Clear SAYT requests for products if SAYT is hidden

geoffcoutts:
`debouncedRequestSaytProducts()` should be cleared after a user wilfully exits/hides the SAYT window.

### DIFF
--- a/packages/web-components/@sfx/sayt/src/sayt.ts
+++ b/packages/web-components/@sfx/sayt/src/sayt.ts
@@ -207,7 +207,7 @@ export default class Sayt extends LitElement {
   }
 
   /**
-   * Changes the `visible` property to `false`.
+   * Changes the [[visible]] property to `false`.
    * This will clear debounced calls for SAYT products. 
    */
   hideSayt(): void {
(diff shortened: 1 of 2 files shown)


packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -211,6 +211,8 @@ export default class Sayt extends LitElement {
    */
   hideSayt(): void {
     this.visible = false;
+    this.debouncedRequestSaytProducts.clear();
+    this.visible = false;
```
jmbtutor: Why is this necessary? If it's a timing issue, would it be sufficient to clear first, then set visible?

packages/web-components/@sfx/sayt/src/sayt.ts
```diff
@@ -207,10 +207,12 @@ export default class Sayt extends LitElement {
   }
 
   /**
-   * Changes the `visible` property to `false`.
+   * Changes the [[visible]] property to `false`.
+   * This will clear debounced calls for SAYT products.
```
jmbtutor:
```suggestion
   * This will also clear debounced calls for SAYT products.
```

PR #44, opened 2019-10-28
SFX-146: Functional End to End Testing

ssbelle:


.testcaferc.json
```diff
@@ -0,0 +1,4 @@
+{
+    "browsers": ["chrome", "firefox"],
+    "skipJsErrors": true
```
DanielMoniz-GBI: For anyone's interest:

https://devexpress.github.io/testcafe/documentation/using-testcafe/command-line-interface.html#-e---skip-js-errors

@ssbelle Is there a specific reason why you feel we need to do this? We can always remove it down the road, but I'm curious about your intentions here.

demo/demo.html
```diff
@@ -3,8 +3,8 @@
   <head>
     <title>Demo of Logic and View Integration</title>
     <script src="../dist/components.js"></script>
-    <script src="https://cdn.groupbycloud.com/sfx-all-plugins-bundle.js"></script>
-    <!-- <script src="./temp/sfx-all-plugins-bundle.js"></script> -->
+    <!-- <script src="https://cdn.groupbycloud.com/sfx-all-plugins-bundle.js"></script> -->
+    <script src="./temp/sfx-all-plugins-bundle.js"></script>
```
DanielMoniz-GBI: I've been trying to leave this commented out. However, given that the CDN version is pretty old, there is some argument to be made that we should switch to the temp bundle for now until we have a proper release cycle.

README.md
```diff
@@ -82,6 +82,29 @@ import '../../../../../test-tools/setup';
...
+```
+
+- To run the tests for chrome only:
+```sh
+yarn functional:test:chrome
+```
+
+- To run the tests for safari only:
+```sh
+yarn functional:test:safari
```
DanielMoniz-GBI: Comparing to an earlier command,
```
yarn test:interaction
```
the notation thus far is to start with `test`, follow it up with the type of testing (eg. `:interaction`), and then follow that up with anything more specific (eg. `:all`, `:chrome`, etc.). Let's be consistent with our previous test commands.

.testcaferc.json
ssbelle: My memory was that it made safari work without errors, but I've since  upgraded the package version and I don't see that error now. Removed

demo/demo.html
ssbelle: Yes my tests will fail, since no terms or products will populate when using the cdn now. What would you like to do? @DanielMoniz-GBI

test.ts
```diff
@@ -0,0 +1,48 @@
...
+    .pressKey( 'home left c h e c k space d r e s s' )
+});
+
+test('SFX sayt items are hoverable', async t => {
+await t
+    .setTestSpeed(0.4)
+    .hover(searchbox)
+    .click(searchbox)
+    .pressKey( 'home left c h e c k space d r e s s' )
+    .setTestSpeed(0.6)
```
DanielMoniz-GBI: Can you speak to the need to set the test speed, and to do so potentially more than once during a test?

demo/demo.html
DanielMoniz-GBI: I'm good with leaving this as you have it!

test.ts
ssbelle: Sure, I found that certain actions were happening too quickly to register (tp the eye) so I slowed some down to different degrees. I can change of course if you like

README.md
ssbelle: Changed

test-tools/testcafe.ts
```diff
@@ -0,0 +1,73 @@
...
+  // confirm searchbox exists
+  await t.expect(searchbox).ok('Searchbox exists')
+  await t.hover(searchbox)
+  await t.click(searchbox)
+  // confirm searchbox input
+  await t.typeText(searchbox, 'check dress')
+  await t.expect(input.value).contains('check dress', 'Searchbox has successful input value')
+  // confirm user can hover over autocomplete terms
+  const autocompleteTermIndex = 2;
+  const thirdAutocompleteTerm = await autocomplete.child('ul').child('li').nth(autocompleteTermIndex)();
```
DanielMoniz-GBI: I think you can just call this constant `autocompleteTerm`, unless you disagree.

test-tools/testcafe.ts
```diff
@@ -0,0 +1,73 @@
...
+  await t.click(searchbox)
+  // confirm searchbox input
+  await t.typeText(searchbox, 'check dress')
+  await t.expect(input.value).contains('check dress', 'Searchbox has successful input value')
+  // confirm clear button exists
+  await t.expect(clearButton).ok('Clear button exists')
+  // confirm click of clear button clears searchbox input
+  await t.hover(clearButton)
+  await t.click(clearButton)
+  await t.expect(input.value).contains('', 'Searchbox has successfully been cleared')
```
DanielMoniz-GBI: I think this is a misuse of `contains`. Checking that a string contains the empty string will produce `true` 100% of the time.

You could try to check that it's equal to the empty string instead. The assertion API seems fairly limited. As a somewhat confusing last resort, you could check that the empty string contains `input.value`, which would prove that they are indeed both empty.

README.md
```diff
@@ -82,6 +82,29 @@ import '../../../../../test-tools/setup';
 
 Keywords from the testing frameworks can then be imported to your test files from `utils.ts`.
 
+### Functional End-to-End Testing
+
+Functional tests are run with [Testcafe](https://devexpress.github.io/testcafe/), which runs tests for each browser environment configured. The following browsers are currently tested:
+
+- Chrome
+- Firefox
```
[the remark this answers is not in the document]
DanielMoniz-GBI: @ssbelle Eventually, but I think this ticket is fulfilled. We can just work to wrap it up.

You can add a comment to the ticket in Jira if you have items that you think should be done in the future.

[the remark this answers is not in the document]
ssbelle: For the record: 

> @Daniel Moniz @Chantal https://devexpress.github.io/testcafe/documentation/extending-testcafe/browser-provider-plugin/#specifying-a-browser-for-a-test-task its definitely possible, we would have to decide what versions we want to cover(by listing in the config) and instead of using browsers on our locals possibly direct our testcafe code to use a service that has all those browser versions

PR #47, opened 2019-11-04
SFX-421: Update demo to handle area and collection in search requests

geoffcoutts:
Add `area` and `collection` as attributes to one of the search boxes in the demo to include them as search request configurations. Also updated the search product transformer to now parse prices from the results of the new collection.

### DIFF
--- a/demo/demo.html
+++ b/demo/demo.html
@@ -36,7 +36,7 @@
   </head>
 
   <body>
-    <gbe-search-box id="main-search" placeholder="search" clearbutton searchbutton></gbe-search-box>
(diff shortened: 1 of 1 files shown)


geoffcoutts:
> I'm inclined to leave out the bundle changes as well.

@jmbtutor I put the bundle changes in because I noticed that it was updated recently in another PR. Doesn't it make sense to also keep the bundle up to date so that any new changes from logic are reflected in this internal demo?

jmbtutor:
@geoffcoutts If I remember correctly, we let that one in since at the time, the bundle was horribly out of date, but we made note that we probably shouldn't be doing it too often because it bloats the repo and because we're hopefully going to make proper releases soon.

PR #50, opened 2019-11-08
ELE-383: Create storybook events logger panel

ssbelle:
Uses a custom package (storybook-events-logger) to log each registered event in a story to the Events Logger panel for inspection.
Storybook-events-logger logs the name and payload of the event but can be modified easily to include other attributes.

To use simply add storybook-events-logger as package. Import the logEvent method from the pacakage, and pass an event to logEvent ex: `logEvent(event)`.

NPM package can be found here: https://www.npmjs.com/package/storybook-events-logger
github repo here for now: https://github.com/ssbelle/storybook-events-logger


packages/web-components/@groupby/elements-search-box/stories/index.ts
```diff
@@ -11,6 +18,20 @@ const updateTextEvent = [
...
+const eventsToListen = [
+  SEARCHBOX_CLEAR,
+  SEARCHBOX_CLICK,
+  SEARCHBOX_INPUT,
+  SEARCH_REQUEST,
+  UPDATE_SEARCH_TERM
+];
+
+eventsToListen.forEach((eventName) => {
+  addEventListener(eventName, (event) => {
```
jmbtutor: We can pass the `logEvent` function directly as the callback:
```js
addEventListener(eventName, logEvent);
```

packages/web-components/@groupby/elements-search-box/stories/index.ts
```diff
@@ -11,6 +18,20 @@ const updateTextEvent = [
...
 
+const eventsToListen = [
+  SEARCHBOX_CLEAR,
+  SEARCHBOX_CLICK,
+  SEARCHBOX_INPUT,
+  SEARCH_REQUEST,
+  UPDATE_SEARCH_TERM
+];
+
+eventsToListen.forEach((eventName) => {
```
jmbtutor: If we never use `eventsToListen` again, you can simply inline it and invoke the `forEach` directly on the array, unless you think the way you have it is much clearer.

ssbelle: @jmbtutor I was actually thinking that we would use this function again in other stories and maybe I should move it to some appropriate utils place maybe in the storybook config folder?

DanielMoniz-GBI: @ssbelle I also think it's likely we'll re-use it, and so a more global function is definitely useful (as you have implemented). Another consideration is whether this function should exist in your library so we can simply import it. It's not necessary right away, but it's worth considering if your library's purpose extends to that.

packages/web-components/@groupby/elements-search-box/stories/index.ts
```diff
@@ -11,6 +17,16 @@ const updateTextEvent = [
...
 
+const eventsToListen = [
+  SEARCHBOX_CLEAR,
+  SEARCHBOX_CLICK,
+  SEARCHBOX_INPUT,
+  SEARCH_REQUEST,
+  UPDATE_SEARCH_TERM
+];
+
+addStorybookListeners(eventsToListen);
```
DanielMoniz-GBI: @ssbelle Johann's comment about inlining this still applies here. If you're not going to use `eventsToListen` again, you can simply do:
```suggestion
addStorybookListeners([
  SEARCHBOX_CLEAR,
  SEARCHBOX_CLICK,
  SEARCHBOX_INPUT,
  SEARCH_REQUEST,
  UPDATE_SEARCH_TERM,
]);
```

packages/web-components/@groupby/elements-search-box/stories/index.ts
ssbelle: @DanielMoniz-GBI I currently moved it to the storybook common js file with other utils functions we use throughout multiple stories. Does that work?

.storybook/common.ts
```diff
@@ -186,3 +187,7 @@ export interface StorybookCustomEvent {
   name: string;
   payload: any;
 }
+
+export function addStorybookListeners(events) {
+  events.forEach(e => addEventListener(e, logEvent));
```
jmbtutor: I don't think the linter runs in this file, but we should be consistent anyway.

```suggestion
  events.forEach((e) => addEventListener(e, logEvent));
```

packages/web-components/@groupby/elements-search-box/stories/index.ts
```diff
@@ -11,6 +17,14 @@ const updateTextEvent = [
   },
 ];
 
+addStorybookListeners([
+  SEARCHBOX_CLEAR,
+  SEARCHBOX_CLICK,
+  SEARCHBOX_INPUT,
+  SEARCH_REQUEST,
+  UPDATE_SEARCH_TERM
```
geoffcoutts: Please use trailing commas.
```suggestion
  UPDATE_SEARCH_TERM,
```

packages/web-components/@groupby/elements-search-box/stories/index.ts
```diff
@@ -1,7 +1,13 @@
 import { storiesOf } from '@storybook/html';
 import { withKnobs, text, boolean } from '@storybook/addon-knobs';
-import { UPDATE_SEARCH_TERM } from '@groupby/elements-events';
-import { getDisplayCode } from '../../../../../.storybook/common';
+import {
+  SEARCHBOX_CLEAR,
+  SEARCHBOX_CLICK,
+  SEARCHBOX_INPUT,
+  SEARCH_REQUEST,
+  UPDATE_SEARCH_TERM
```
geoffcoutts: Please use trailing commas. 
```suggestion
  UPDATE_SEARCH_TERM,
```

.storybook/common.ts
geoffcoutts: I'm assuming that this is attaching to the global window by default? Thoughts on being more explicit about this?

packages/web-components/@groupby/elements-search-box/stories/index.ts
ssbelle: @DanielMoniz-GBI yes I had thought about moving that functionality into the package, I can definitely do that now or later

DanielMoniz-GBI: @ssbelle I'm guessing it'll happen later!

PR #52, opened 2019-11-11
ELE-195 - Target ES5

ChantalDesrochers:
- babel-loader is used to convert code into a backwards compatible version of JavaScript (to run in IE11)
- core-js, webcomponents-bundle, and custom-elements-es5-adapter used for polyfilling
- demo.html file modified to be compatible with IE11 (removal of arrow functions, spread, etc.)
- Karma and Storybook use ts-loader (es2015)


demo/demo.js
```diff
@@ -6,23 +6,22 @@ function parseRecord(record) {
...
-      imageSrc: v.productImage,
-      imageAlt: v.description,
-    },
-  }));
+  return variants.map(function(v) {
+    return {
+      color: v.standardColor,
+      text: v.standardColor,
+      image: v.swatchImage,
+      product: {imageSrc: v.productImage, imageAlt: v.description, },
```
DanielMoniz-GBI: Jus a space missing here:
```suggestion
      product: { imageSrc: v.productImage, imageAlt: v.description, },
```

demo/demo.js
```diff
@@ -6,23 +6,22 @@ function parseRecord(record) {
   if (!nonvisualVariants[0]) throw new Error('No nonvisual variants');
 
   return {
-    data,
-    firstVariant,
-    nonvisualVariants,
-    visualVariants,
+    data: data,
```
ssbelle: why did we break this out, does the es6 shorthand not work here?

PR #55, opened 2019-11-15
ELE-453: Cache products on initial load

ssbelle:


demo/temp/gbe-all-plugins-bundle.js
```diff
@@ -14907,6 +14907,7 @@ var CacheDriverPlugin = /** @class */ (function () {
      * callbacks.
```
jmbtutor: Please revert this file. Since we're getting a release process soon, we can put up with this not being up to date for a little bit longer.

packages/web-components/@groupby/elements-products/src/products-base.ts
```diff
@@ -40,6 +45,34 @@ export default class ProductsBase extends Base {
...
+  }
+
+  /**
+   * Generates a string intended to be used as the name of the return event in
+   * cache requests for the component.
+   *
+   * @param componentName The name of the component for which to generate the return event name.
+   * @param componentId A random string that acts as a unique identifier for the component.
+   */
+  getCacheResponseEventName(componentName: string, componentId: string): string {
```
jmbtutor: This can be turned into a getter. Please refer to the autocomplete component for an example.

packages/web-components/@groupby/elements-products/src/products-sayt.ts
```diff
@@ -40,6 +57,18 @@ export default class ProductsSayt extends ProductsBase {
...
+  }
+
+  /**
+   * Receives an event for populating initial data.
+   * Intended to be used on mount of this component.
+   *
+   * @param event The event object.
+   */
+  setProductsFromCacheEvent(event: CustomEvent<CacheResponsePayload>): void {
+    const data = event.detail.data || {};
```
jmbtutor: @DanielMoniz-GBI: I would normally suggest the refactor below, but it does change the behaviour slightly (defaulting only on undefined rather than any falsy value). Would that be acceptable in this case?  

```suggestion
    const { data = {} } = event.detail;
```

@ssbelle If you copied this code from autocomplete, then ignore this suggestion for consistency.

packages/web-components/@groupby/elements-products/src/products-search.ts
```diff
@@ -15,13 +17,26 @@ import { ProductsBase } from '.';
...
+   * A random string suitable for use in stable IDs related to this
+   * component.
+   */
+  protected componentId = shortid.generate();
+
+  /**
+   * A string intended to be used as the name of the return event in
+   * cache requests for this component.
+   */
+  protected cacheResponseEventName = this.getCacheResponseEventName('products-search', this.componentId);
```
jmbtutor: Same feedback as in earlier components.

packages/web-components/@groupby/elements-products/test/unit/products.test.ts
```diff
@@ -55,6 +60,31 @@ describe('Products Base Component', () => {
...
+  describe('requestInitialData', () => {
+    it('should dispatch a cache request event', () => {
+      const dispatchElementsEvent = stub(component, 'dispatchElementsEvent');
+      const name = 'event';
+      const group = 'group';
+      const returnEvent = 'event-return';
+
+      component.requestInitialData(name, group, returnEvent);
+
+      expect(dispatchElementsEvent).to.have.been.calledWith(CACHE_REQUEST, { name, group, returnEvent });
```
jmbtutor: I think we've been using the simple `to.be.calledWith` elsewhere.

```suggestion
      expect(dispatchElementsEvent).to.be.calledWith(CACHE_REQUEST, { name, group, returnEvent });
```

packages/web-components/@groupby/elements-products/test/unit/products.test.ts
```diff
@@ -67,26 +97,66 @@ describe('Products Sayt Component', () => {
   itShouldExtendClass(() => component, ProductsBase);
 
   describe('connectedCallback', () => {
+    let cacheResponseEventName;
+
+    beforeEach(() => {
+      cacheResponseEventName = component.cacheResponseEventName = 'sayt-cache-response';
```
jmbtutor: If you take the getter approach, you can use the getter directly without having to stub it.

For the record, if you were to keep this, it would be better style to make `cacheResponseEventName` a `const` and simply set `component.cacheResponseEventName = cacheResponseEventName` here.

packages/web-components/@groupby/elements-products/test/unit/products.test.ts
```diff
@@ -67,26 +97,66 @@ describe('Products Sayt Component', () => {
...
+    });
+
     itShouldCallParentMethod(() => component, 'connectedCallback');
 
-    it('should set up an event listener for a sayt products event to set products', () => {
-      const addEventListener = sinon.stub(window, 'addEventListener');
+    it('should add event listeners for sayt products events', () => {
+      const addEventListener = stub(window, 'addEventListener');
```
jmbtutor: This function will always add event listeners, so we should stub this in `beforeEach` so we don't leak listeners.

packages/web-components/@groupby/elements-products/test/unit/products.test.ts
```diff
@@ -152,26 +222,56 @@ describe('Products Search Component', () => {
   itShouldExtendClass(() => component, ProductsBase);
 
   describe('connectedCallback', () => {
+    let cacheResponseEventName;
+
+    beforeEach(() => {
+      cacheResponseEventName = component.cacheResponseEventName = 'sayt-cache-response';
```
jmbtutor: Same feedback as the similar line.

packages/web-components/@groupby/elements-products/test/unit/products.test.ts
```diff
@@ -152,26 +222,56 @@ describe('Products Search Component', () => {
...
+      expect(requestInitialData).to.be.calledWith(SEARCH_RESPONSE, component.group, cacheResponseEventName);
     });
   });
 
   describe('disconnectedCallback', () => {
     itShouldCallParentMethod(() => component, 'disconnectedCallback');
 
-    it('should remove an event listener for search responses', () => {
+    it('should remove event listeners for sayt products', () => {
       const removeEventListener = sinon.stub(window, 'removeEventListener');
```
jmbtutor: Since you're modifying the code around it anyway, we may as well normalize this now.
```suggestion
      const removeEventListener = stub(window, 'removeEventListener');
```

packages/web-components/@groupby/elements-products/src/products-base.ts
```diff
@@ -40,6 +45,34 @@ export default class ProductsBase extends Base {
...
+   * @param group The name of the event group that the component belongs to.
+   * @param returnEvent The name of the event under which to dispatch the cached data.
+   *
+   */
+  requestInitialData(name: string, group: string, returnEvent: string): void {
+    const payload: CacheRequestPayload = {
+      name,
+      group,
+      returnEvent,
+    };
```
DanielMoniz-GBI: This can probably just be done on one line.

packages/web-components/@groupby/elements-products/src/products-sayt.ts
```diff
@@ -15,13 +17,26 @@ import { ProductsBase } from '.';
  */
 @customElement('gbe-products-sayt')
 export default class ProductsSayt extends ProductsBase {
+  /**
+   * A random string suitable for use in stable IDs related to this
+   * component.
+   */
+  protected componentId = shortid.generate();
```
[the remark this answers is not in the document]
DanielMoniz-GBI: There is a story for it. We may even do it in Phase 1; possibly this sprint!

packages/web-components/@groupby/elements-products/src/products-sayt.ts
```diff
@@ -15,13 +17,26 @@ import { ProductsBase } from '.';
...
+   */
+  protected cacheResponseEventName = this.getCacheResponseEventName('products-sayt', this.componentId);
+
   /**
    * Binds relevant methods.
    */
   constructor() {
     super();
 
     this.setProductsFromEvent = this.setProductsFromEvent.bind(this);
```
jmbtutor: To make this name less ambiguous, consider renaming this function to `setProductsFromProductsEvent` so it can't be confused with `setProductsFromCacheEvent`.

packages/web-components/@groupby/elements-products/src/products-sayt.ts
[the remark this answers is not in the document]
ssbelle: This threw errors in linter - come back to

packages/web-components/@groupby/elements-products/src/products-base.ts
```diff
@@ -40,6 +45,34 @@ export default class ProductsBase extends Base {
     }
   }
 
+  /**
+   * Requests initial data for the component.
+   *
+   * @param name The name of the initial cached data to return.
+   * @param group The name of the event group that the component belongs to.
+   * @param returnEvent The name of the event under which to dispatch the cached data.
+   *
```
DanielMoniz-GBI: This line can be removed.

packages/web-components/@groupby/elements-products/src/products-sayt.ts
DanielMoniz-GBI: I'm assuming we'd want to default on `undefined` and maybe `null`. We probably don't want to default on any other value.

packages/web-components/@groupby/elements-products/src/products-sayt.ts
```diff
@@ -15,13 +17,26 @@ import { ProductsBase } from '.';
...
+   * A random string suitable for use in stable IDs related to this
+   * component.
+   */
+  protected componentId = shortid.generate();
+
+  /**
+   * A string intended to be used as the name of the return event in
+   * cache requests for this component.
+   */
+  protected cacheResponseEventName = this.getCacheResponseEventName('products-sayt', this.componentId);
```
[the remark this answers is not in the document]
ssbelle: done

packages/web-components/@groupby/elements-products/src/products-base.ts
ssbelle: @jmbtutor I was not able to make this a getter, it threw an error about `A 'get' accessor cannot have parameters.` I did follow your suggestion to make both the protected cacheResponseEventName in products-search and products-sayt into getters. I'll submit this next round momentarily and see what you think. I think there might be a cleaner way to do all this but Im interested in your next review

packages/web-components/@groupby/elements-products/src/products-sayt.ts
ssbelle: @jmbtutor I did copy this code essentially from the autocomplete example here: https://github.com/groupby/elements-view/pull/45/files

packages/web-components/@groupby/elements-products/src/products-search.ts
ssbelle: done

packages/web-components/@groupby/elements-products/test/unit/products.test.ts
ssbelle: @jmbtutor it seems to me that the addListeners and RemoveListeners are set at the correct scope/place within products-sayt and products-search tests. They are not currently used in multiple tests within the connected and disconnected callbacks and so were set within the one test that uses them respectively. That seems correct to me but are you suggesting that addListener should be set in a beforeEach for the connectedCallback even though it is not used in multiple tests within the connectedCallback section? Or some other higher level beforeEach? Sorry Im just wanting to make sure I understand your feedback, and I'm unsure if setting these listeners when they arent used is correct? Thanks!

packages/web-components/@groupby/elements-products/src/products-sayt.ts
```diff
@@ -39,15 +50,35 @@ export default class ProductsSayt extends ProductsBase {
...
+    window.removeEventListener(SAYT_PRODUCTS_RESPONSE, this.setProductsFromProductsEvent);
+    window.removeEventListener(this.cacheResponseEventName, this.setProductsFromCacheEvent);
+  }
+
+  /**
+   * A string intended to be used as the name of the return event in
+   * cache requests for this component.
+   */
+  get cacheResponseEventName(): string {
+    return this.getCacheResponseEventName('products-sayt', this.componentId);
```
ssbelle: @jmbtutor this is the line I am confused about. This is how I interpreted your suggestion to make getters, however maybe there is a better way to do this?

packages/web-components/@groupby/elements-products/src/products-sayt.ts
ssbelle: Did you want me to change this? Or leave the same, as it is the same as autocomplete @jmbtutor @DanielMoniz-GBI

packages/web-components/@groupby/elements-products/src/products-sayt.ts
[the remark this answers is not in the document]
ssbelle: done

packages/web-components/@groupby/elements-products/src/products-sayt.ts
ssbelle: leave

packages/web-components/@groupby/elements-products/src/products-base.ts
ssbelle: done

packages/web-components/@groupby/elements-products/src/products-sayt.ts
ssbelle: done

packages/web-components/@groupby/elements-products/test/unit/products.test.ts
ssbelle: explained and done

packages/web-components/@groupby/elements-products/src/products-base.ts
```diff
@@ -40,6 +52,26 @@ export default class ProductsBase extends Base {
...
+  }
+
+  /**
+   * Generates a string intended to be used as the name of the return event in
+   * cache requests for the component.
+   */
+  get cacheResponseEventName(): string {
+    return `${CACHE_RESPONSE_PREFIX}${this.tagName}-${this.componentId}`;
+  }
+
```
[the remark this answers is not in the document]
ssbelle: @geoffcoutts even though this would mean duplication of code? Until we have another solution for where to house the cache methods isn't a shared class the best place? @jmbtutor @DanielMoniz-GBI @ChantalDesrochers  My ticket indicated for me to try to leverage the base class but I will move it if you feel strongly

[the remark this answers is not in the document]
ssbelle: @geoffcoutts also do you mean both the cacheResponseEventName and the requestInitialData methods (those are both in base but used in products search and sayt)? For clarity is your suggestion to move those both out into the separate files?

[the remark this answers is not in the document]
DanielMoniz-GBI: @geoffcoutts @ssbelle The `ProductsBase` class is not currently related to events, but I don't think its purpose was to avoid events. The reason we have multiple Products classes (that inherit from Base) was to solve a specific problem: how can we allow for product grids that:
a) respond to different events,
b) are maintainable by the developers (us), and
c) are clear for and easy to use by clients?

As far as I can see, if we can avoid duplication between the product grids, we should do so by adding code to `ProductsBase`. If the logic required to do so is complex, we should duplicate it across the separate components (as we did with the events). In this case, the logic is generalizable and seems appropriate for the base class.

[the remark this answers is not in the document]
jmbtutor: On further reflection, I agree that utilities for the events can be in `ProductsBase` (and really, this particular getter should probably live in `Base`) as long as `ProductsBase` remains flexible and doesn't require its descendants to send or handle specific events. Having this getter here is compatible with that.

packages/web-components/@groupby/elements-products/test/unit/products.test.ts
```diff
@@ -12,10 +16,13 @@ import ProductsSayt from '../../src/products-sayt';
 import ProductsSearch from '../../src/products-search';
 
 describe('Products Base Component', () => {
+  const cacheResponseEventName = 'cache-response-event-name';
+  let cacheResponseEventNameStub;
```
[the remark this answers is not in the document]
ssbelle: removed from top level

packages/web-components/@groupby/elements-products/test/unit/products.test.ts
```diff
@@ -55,6 +62,33 @@ describe('Products Base Component', () => {
...
+    });
+  });
+
+  describe('cacheResponseEventName', () => {
+    it('should return an event name for receiving cached data', () => {
+      const componentId = component.componentId = 'some-id';
+      const tagName = 'some-products-component';
+      const expectedCacheResponseEventName = `${CACHE_RESPONSE_PREFIX}${tagName}-${componentId}`;
+      stub(component, 'tagName').get(() => tagName);
+      cacheResponseEventNameStub.restore();
```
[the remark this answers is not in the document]
ssbelle: removed restore, test passes

packages/web-components/@groupby/elements-products/src/products-base.ts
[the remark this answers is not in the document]
geoffcoutts: @ssbelle @DanielMoniz-GBI @jmbtutor All good. I didn't have any major problems with it going in `ProductsBase`, it just presently didn't conform to what the class was doing. Given that these methods are going to be staying put in `ProductsBase`, does it make sense to also move the response handler there as well?

packages/web-components/@groupby/elements-products/test/unit/products.test.ts
```diff
@@ -101,7 +176,7 @@ describe('Products Sayt Component', () => {
 
     it('should set products to an empty array if the event payload does not contain products', () => {
       const event = { detail: {} };
-      component.setProductsFromEvent(event);
+      component.setProductsFromProductsEvent(event);
```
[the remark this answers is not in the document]
ssbelle: changed

packages/web-components/@groupby/elements-products/test/unit/products.test.ts
```diff
@@ -1,7 +1,11 @@
-import { SAYT_PRODUCTS_RESPONSE, SEARCH_RESPONSE } from '@groupby/elements-events';
+import {
+  CACHE_RESPONSE_PREFIX,
+  CACHE_REQUEST,
+  SAYT_PRODUCTS_RESPONSE,
+  SEARCH_RESPONSE,
+} from '@groupby/elements-events';
```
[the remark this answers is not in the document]
ssbelle: fixed

packages/web-components/@groupby/elements-products/src/products-search.ts
```diff
@@ -2,6 +2,7 @@ import { css, CSSResult, customElement } from 'lit-element';
 import {
   SEARCH_RESPONSE,
   Product,
+  CacheResponsePayload,
   SearchResponsePayload,
 } from '@groupby/elements-events';
 import { ProductsBase } from '.';
```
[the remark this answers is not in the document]
ssbelle: fixed

packages/web-components/@groupby/elements-products/src/products-base.ts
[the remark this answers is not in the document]
ssbelle: leave

packages/web-components/@groupby/elements-products/src/products-sayt.ts
```diff
@@ -2,6 +2,7 @@ import { css, CSSResult, customElement } from 'lit-element';
 import {
   SAYT_PRODUCTS_RESPONSE,
   Product,
+  CacheResponsePayload,
   SaytProductsResponsePayload,
 } from '@groupby/elements-events';
```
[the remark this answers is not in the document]
ssbelle: fixed

packages/web-components/@groupby/elements-products/src/products-base.ts
```diff
@@ -6,7 +6,13 @@ import {
   property,
   TemplateResult,
 } from 'lit-element';
-import { Product } from '@groupby/elements-events';
+import {
+  CACHE_RESPONSE_PREFIX,
+  CACHE_REQUEST,
```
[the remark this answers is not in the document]
ssbelle: fixed

packages/web-components/@groupby/elements-products/test/unit/products.test.ts
```diff
@@ -67,30 +101,71 @@ describe('Products Sayt Component', () => {
...
+
+      expect(removeEventListener).to.be.calledWith(
+        component.cacheResponseEventName,
+        component.setProductsFromCacheEvent
+      );
+    });
+  });
+
+  describe('setProductsFromCacheEvent', () => {
+    it('should set products to an empty array if the event payload does not contain products', () => {
```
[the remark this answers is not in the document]
ssbelle: i included the case where it does contain products

packages/web-components/@groupby/elements-products/src/products-base.ts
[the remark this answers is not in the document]
ssbelle: done

PR #56, opened 2019-11-20
ELE-480: Add data initialization check to cache requests

geoffcoutts:
An `_initialized` boolean property was added to `Base` with the intention of being set to true once a component has received data either through a search call or attribute value. This value is checked when reacting to cache responses on whether to use the cache data. 

An accompanying decorator was added to extend a property's set accessor to also set a component's `_initialized` value to true.

### DIFF
--- a/packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
+++ b/packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
@@ -82,8 +82,11 @@ export default class Autocomplete extends Base {
     super.connectedCallback();
 
     window.addEventListener(AUTOCOMPLETE_RESPONSE, this.receivedResults);
-    window.addEventListener(this.initialDataResponseEventName, this.receiveInitialData);
-    this.requestInitialData();
+
+    if (!this._initialized) {
+      window.addEventListener(this.initialDataResponseEventName, this.receiveInitialData);
+      this.requestInitialData();
+    }
 
     const role = this.getAttribute('role');
     const roles = role ? role.split(' ') : [];
--- a/packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
+++ b/packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
@@ -181,13 +181,17 @@ export default class Autocomplete extends Base {
 
   /**
    * Receives an event for populating initial data.
+   * Will not populate if the component has already received data
+   * from another source.
    * Intended to be used on mount of this component.
    *
    * @param event The event object.
    */
   receiveInitialData(event: CustomEvent<CacheResponsePayload>): void {
     const data = event.detail.data || {};
-    this.results = data.results || [];
+    if (!this._initialized) {
+      this.results = data.results || [];
+    };
   }
 
   /**
--- a/packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
+++ b/packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
@@ -35,9 +35,7 @@ export default class Autocomplete extends Base {
   /**
    * Autocomplete request results.
    */
-  @dataInitializer('_initialized')
-  @property({ type: Array })
-  results: AutocompleteResultGroup<AutocompleteSearchTermItem>[] = [];
+  @dataInitializer() @property({ type: Array })results: AutocompleteResultGroup<AutocompleteSearchTermItem>[] = [];
 
   /**
    * The text to use in the header.
--- a/packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
+++ b/packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
@@ -123,22 +123,9 @@ export default class Autocomplete extends Base {
    */
   disconnectedCallback(): void {
     super.disconnectedCallback();
+    console.log('in disconnectedCallback')
     window.removeEventListener(AUTOCOMPLETE_RESPONSE, this.receivedResults);
     window.removeEventListener(this.initialDataResponseEventName, this.receiveInitialData);
-  }
-
-  /**
-   * Synchronizes property values when attributes change.
-   *
-   * @param name Name of the property.
-   * @param oldVal The old value of the property.
-   * @param newVal The new value of the property.
-   */
-  attributeChangedCallback(name: string, oldVal: string | null, newVal: string | null): void {
-    super.attributeChangedCallback(name, oldVal, newVal);
-    if (name === 'results') {
-    //   this._initialized = true;
-    }
   }
 
   /**
@@ -146,16 +133,12 @@ export default class Autocomplete extends Base {
    * The changes to the following properties are listened for:
    *
    * - `selectedIndex`
-   * - `results`
    *
    * @param changedProps A map of the all the changed properties.
    */
   updated(changedProps: PropertyValues): void {
     if (changedProps.has('selectedIndex')) {
       this.dispatchSelectedTerm();
-    }
-    if (changedProps.has('results')) {
-      this._initialized = true;
     }
   }
 
--- a/packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
+++ b/packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
@@ -136,7 +136,7 @@ export default class Autocomplete extends Base {
     super.attributeChangedCallback(name, oldVal, newVal);
     if (name === 'results') {
       this._initialized = true;
-    };
+    }
   }
 
   /**
@@ -194,7 +194,7 @@ export default class Autocomplete extends Base {
     const data = event.detail.data || {};
     if (!this._initialized) {
       this.results = data.results || [];
-    };
+    }
   }
 
   /**
--- a/packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
+++ b/packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
@@ -85,7 +85,6 @@ export default class Autocomplete extends Base {
 
     window.addEventListener(AUTOCOMPLETE_RESPONSE, this.receivedResults);
 
-    // eslint-disable-next-line no-underscore-dangle
     if (!this._initialized) {
       window.addEventListener(this.initialDataResponseEventName, this.receiveInitialData);
       this.requestInitialData();
@@ -181,7 +180,6 @@ export default class Autocomplete extends Base {
    */
   receiveInitialData(event: CustomEvent<CacheResponsePayload>): void {
     const data = event.detail.data || {};
-    // eslint-disable-next-line no-underscore-dangle
     if (!this._initialized) {
       this.results = data.results || [];
     }
--- a/packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
+++ b/packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
@@ -144,12 +144,16 @@ export default class Autocomplete extends Base {
    * The changes to the following properties are listened for:
    *
    * - `selectedIndex`
+   * - `results`
    *
    * @param changedProps A map of the all the changed properties.
    */
   updated(changedProps: PropertyValues): void {
     if (changedProps.has('selectedIndex')) {
       this.dispatchSelectedTerm();
+    }
+    if (changedProps.has('results')) {
+      this._initialized = true;
     }
   }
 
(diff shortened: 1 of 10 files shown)


packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
```diff
@@ -35,7 +35,7 @@ export default class Autocomplete extends Base {
   /**
    * Autocomplete request results.
    */
-  @property({ type: Array }) results: AutocompleteResultGroup<AutocompleteSearchTermItem>[] = [];
+  @dataInitializer() @property({ type: Array })results: AutocompleteResultGroup<AutocompleteSearchTermItem>[] = [];
```
jmbtutor: This is missing a space.
```suggestion
  @dataInitializer() @property({ type: Array }) results: AutocompleteResultGroup<AutocompleteSearchTermItem>[] = [];
```

We should consider reformatting our decorator calls in the future to have the property they decorate on a separate line from the decorators. Maybe also have each decorator on separate lines as well.

packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
```diff
@@ -167,11 +170,14 @@ export default class Autocomplete extends Base {
 
   /**
    * Receives an event for populating initial data.
+   * Will not populate if the component has already received data
+   * from another source.
```
jmbtutor: Consider "This function will do nothing if the component had previously received data." At minimum, though, "will not populate" does not read well: I think you'd have to say "This function will not populate the component with data if ..."

packages/web-components/@groupby/elements-autocomplete/test/unit/autocomplete.test.ts
```diff
@@ -92,6 +92,18 @@ describe('Autcomplete Component', () => {
       expect(requestInitialData).to.be.called;
     });
 
+    it('should not request initial data if the component is already initialized', () => {
+      autocomplete._initialized = true;
+
+      autocomplete.connectedCallback();
+
+      expect(windowAddEventListener).to.not.be.calledWith(
```
jmbtutor: Does this not fit on one line? That said, consistency with surrounding code says that this should stay. No action needed.

packages/web-components/@groupby/elements-base/src/base.ts
```diff
@@ -15,3 +20,30 @@ export default abstract class Base extends LitElement {
...
+
+/**
+ * A property decorator that will set the data initialization state in a class.
+ * It will extend the property's setter to set the initialization state to true.
+ * This expects that the property already has defined accessors and a default value.
+ *
+ * @param initProperty The name of the initialization property. Has a default value
+ * of "_initialized".
+ */
+export function dataInitializer(initProperty: string = '_initialized'): PropertyDecorator {
```
jmbtutor: This is a separate entity and not coupled to `Base`, so I recommend splitting this out into a separate file.

packages/web-components/@groupby/elements-base/src/base.ts
```diff
@@ -15,3 +20,30 @@ export default abstract class Base extends LitElement {
...
   }
 }
+
+/**
+ * A property decorator that will set the data initialization state in a class.
+ * It will extend the property's setter to set the initialization state to true.
+ * This expects that the property already has defined accessors and a default value.
+ *
+ * @param initProperty The name of the initialization property. Has a default value
+ * of "_initialized".
```
jmbtutor: Typedoc will take care of pointing out the default value, so you don't have to.

packages/web-components/@groupby/elements-base/src/base.ts
```diff
@@ -15,3 +20,30 @@ export default abstract class Base extends LitElement {
     return this.dispatchEvent(eventToDispatch);
   }
 }
+
+/**
+ * A property decorator that will set the data initialization state in a class.
```
jmbtutor:
```suggestion
 * An accessor decorator that will set the data initialization state in a class.
```

packages/web-components/@groupby/elements-base/src/base.ts
```diff
@@ -15,3 +20,30 @@ export default abstract class Base extends LitElement {
...
+ * It will extend the property's setter to set the initialization state to true.
+ * This expects that the property already has defined accessors and a default value.
+ *
+ * @param initProperty The name of the initialization property. Has a default value
+ * of "_initialized".
+ */
+export function dataInitializer(initProperty: string = '_initialized'): PropertyDecorator {
+  // eslint-disable-next-line func-names
+  return function(target: object, propertyName: string): void {
```
jmbtutor: This can be an arrow function since you don't need a reference to `this`.

packages/web-components/@groupby/elements-base/src/base.ts
```diff
@@ -15,3 +20,30 @@ export default abstract class Base extends LitElement {
...
+    const oldDescriptor = Object.getOwnPropertyDescriptor(target, propertyName);
+    const instanceMap = new WeakMap();
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    function decoratedSetter(newVal: any): void {
```
jmbtutor: I personally don't have a problem with this, but most will recommend the following form instead.

```suggestion
    const decoratedSetter = function(newVal: any): void {
```

packages/web-components/@groupby/elements-base/test/unit/base.test.ts
```diff
@@ -1,5 +1,86 @@
-import { expect, stub } from '../utils';
+import { expect, stub, spy } from '../utils';
```
jmbtutor: Do we have a consistent order for these? If we do, it'll compress better.

packages/web-components/@groupby/elements-base/test/unit/base.test.ts
```diff
@@ -1,5 +1,86 @@
...
+      },
+      set results(val) {
+        this._results = val;
+      },
+      _results: ['a', 'b'],
+    };
+  });
+
+  it('should extend a property setter', () => {
+    const originalSet = spy(testObj, 'results', ['set']);
```
jmbtutor: Another approach is to have empty spies for the getter and setter instead of relying on the functionality of the underlying getter/setter. Have the original getter/setter call the spies and assert that the spies were called in the test.

packages/web-components/@groupby/elements-base/test/unit/base.test.ts
```diff
@@ -1,5 +1,86 @@
...
+
+    testObj.results = ['c', 'd'];
+    testObj.results = ['e', 'f'];
+
+    expect(testObj.init).to.be.true;
+  });
+
+  it('should store the initialization states between multiple instances of the component', () => {
+    const testObj2 = { ...testObj };
+    Object.defineProperty(
```
jmbtutor: What this statement does isn't clear at first glance, so I would add a comment explaining what the purpose of this statement is. If there's a clearer way to write it, that would be even better.

packages/web-components/@groupby/elements-base/test/unit/base.test.ts
```diff
@@ -1,5 +1,86 @@
...
+      testObj2,
+      testPropertyName,
+      Object.getOwnPropertyDescriptor(testObj, 'results')
+    );
+    dataInitializer('init')(testObj, testPropertyName);
```
jmbtutor: `'init'` can go into a variable and be reused in the four instances where it's used.

packages/web-components/@groupby/elements-base/test/unit/base.test.ts
```diff
@@ -1,5 +1,86 @@
...
+  it('should not set the initialize property to true on the first set call', () => {
+    dataInitializer('init')(testObj, testPropertyName);
+
+    testObj.results = ['c', 'd'];
+    testObj.results = ['e', 'f'];
+
+    expect(testObj.init).to.be.true;
+  });
+
+  it('should store the initialization states between multiple instances of the component', () => {
```
jmbtutor: Either this test or another test should ensure that instances are independent. That is, the initialization state of one component should not leak into the initialization state of the other. This can be as simple as adding a third object in this test.

packages/web-components/@groupby/elements-products/src/products-base.ts
```diff
@@ -27,7 +27,7 @@ export default class ProductsBase extends Base {
   /**
    * The product data to be rendered.
    */
-  @property({ type: Array }) products: Product[] = [];
+  @dataInitializer() @property({ type: Array })products: Product[] = [];
```
jmbtutor: Same as the previous instance.

```suggestion
  @dataInitializer() @property({ type: Array }) products: Product[] = [];
```

packages/web-components/@groupby/elements-products/src/products-base.ts
```diff
@@ -67,10 +67,12 @@ export default class ProductsBase extends Base {
 
   /**
    * Receives an event for populating initial product data.
+   * Will not populate it if the product data has already been initialized.
```
jmbtutor: Same feedback as the similar line from the earlier component.

packages/web-components/@groupby/elements-base/test/unit/base.test.ts
geoffcoutts: I believe this was misread as the expectations at the end show two different `init` states.

jmbtutor: Ah, so it is. My mistake.

packages/web-components/@groupby/elements-autocomplete/test/unit/autocomplete.test.ts
geoffcoutts: I went with consistency on the surrounding lines.

packages/web-components/@groupby/elements-autocomplete/src/autocomplete.ts
geoffcoutts: Added the space. For the comment about line splitting, we talked and both of our initial ideas for attaching the linter was the same. Keeping it how it is for now for consistency with other properties. 

For reference, these are the two options: 
```suggestion
  @dataInitializer() @property({ type: Array })
  results: AutocompleteResultGroup<AutocompleteSearchTermItem>[] = [];
```
```suggestion
  @dataInitializer() 
  @property({ type: Array })
  results: AutocompleteResultGroup<AutocompleteSearchTermItem>[] = [];
```

packages/web-components/@groupby/elements-base/test/unit/base.test.ts
geoffcoutts: As discussed, this is doing that.

packages/web-components/@groupby/elements-base/src/base.ts
geoffcoutts: Moved to a `decorators.ts` file in the same directory.

packages/web-components/@groupby/elements-base/src/base.ts
geoffcoutts: After talking it through we decided to keep it as is.

packages/web-components/@groupby/elements-base/test/unit/base.test.ts
geoffcoutts: I made it consistent with the others. This was autoimported and put there and didn't notice. My b.

packages/web-components/@groupby/elements-base/test/unit/base.test.ts
geoffcoutts: Updated to this pattern. Also involved no longer asserting a change in the `results` property on `testObj`, since technically that shouldn't be done anyways as it is testing something other than the decorator.

packages/web-components/@groupby/elements-base/test/unit/base.test.ts
geoffcoutts: Added a couple of comments to hopefully make this more clear.

packages/web-components/@groupby/elements-base/src/decorators.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * A property decorator that will set the data initialization state in a class.
+ * It will extend the property's setter to set the initialization state to true.
+ * This expects that the property already has defined accessors and a default value.
+ *
+ * @param initProperty The name of the initialization property.
+ */
+// eslint-disable-next-line import/prefer-default-export
```
jmbtutor: This is fine for now. Later on, if we have more decorators, we should consider adding a file-wide disable directive at the top:

```typescript
/* eslint-disable import/prefer-default-export */
```

PR #57, opened 2019-11-22
ELE-436: Populate the searchbox with the selected autocomplete term when the Enter key is pressed

ChantalDesrochers:


packages/web-components/@groupby/elements-sayt/stories/index.ts
```diff
@@ -296,6 +308,17 @@ storiesOf('Components|SAYT', module)
               5. Observe that the autocomplete selection changes.
 
 
+          ### The SAYT component paired with a searchbox will update the search term when an autocomplete term is selected by keydown.
+            * Pressing enter on an autocomplete term will emit an \`${UPDATE_SEARCH_TERM}\` event.
+              * This event will contain the autocomplete term with a flag to not trigger a new search as the payload.
+              * The search-box component will subsequently trigger a search.
```
DanielMoniz-GBI: I feel like this sentence needs an "as normal" or something at the end. I say that because it sounds to some degree as if it contradicts the line above it. Why would it trigger a search if we specified that it shouldn't trigger a search?

packages/web-components/@groupby/elements-sayt/stories/index.ts
```diff
@@ -296,6 +308,17 @@ storiesOf('Components|SAYT', module)
...
+            * To demonstrate in this story:
+              1. Visit the **Custom Events** panel, locate the \`${AUTOCOMPLETE_RESPONSE}\` event.
```
DanielMoniz-GBI:
```suggestion
              1. Visit the **Custom Events** panel and locate the \`${AUTOCOMPLETE_RESPONSE}\` event.
```

packages/web-components/@groupby/elements-sayt/src/sayt.ts
```diff
@@ -225,12 +226,28 @@ export default class Sayt extends Base {
...
+   * Processes a keyboard event from the searchbox component's parent element, in order to update the search term
+   * in the input box.
+   * The `updateSearchTerm` method on autocomplete is called in response to an 'Enter' keyboard event.
+   *
+   * @param event The KeyboardEvent object.
+   */
+  updateSearchboxInputTerm(event: KeyboardEvent): void {
+    if (event.key === 'Enter' && this.nodeInSearchbox(event.target as Node)) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const autocomplete = this.querySelector<any>('[data-gbe-ref="autocomplete"]');
```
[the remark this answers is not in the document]
ChantalDesrochers: I understood it to be casting the return value. I followed the same format as was done on lines 555 and 574 in the same file.
I thought it was the same as something like... this.querySelector('something) as any. That way you could call the methods on autocomplete.
I can do some research to double check though, I'm still iffy on TypeScript.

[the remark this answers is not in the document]
ChantalDesrochers: @DanielMoniz-GBI - I imported Autocomplete and used it as the type. Do we know if there's a reason the other autocomplete query selectors use 'any'?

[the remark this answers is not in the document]
jmbtutor: None of them should be using `any`. `querySelector` takes a generic type parameter that helps type the return type and it defaults to `Element`, which is strictly better than `any`. Here's the [relevant line](https://github.com/microsoft/TypeScript/blob/0c17476d09655c7232134ed311edadfb93924e63/lib/lib.dom.d.ts#L11535) in the DOM types file. Note that the return type is `E | null`, so you should still check that the result isn't `null`. I suspect that's why `any` was put in place, but that's hiding a bug and we shouldn't be doing it. 

Adding the generic type argument to `querySelector` allows you to give a hint to TypeScript that you expect the element returned to be of a certain type, so you can reference that are exclusive to that type (in our case, methods like `updateSearchTerm`).

If you don't need it, don't add it, but in this case, passing `Autocomplete` is probably okay. I wonder if we'll run into any issues if the developers decide to make a partially type-compatible component.

packages/web-components/@groupby/elements-sayt/src/sayt.ts
```diff
@@ -23,6 +23,8 @@ import {
 } from '@groupby/elements-events';
 // eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved
 import { Base } from '@groupby/elements-base';
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved
+import { Autocomplete } from '@groupby/elements-autocomplete';
```
jmbtutor: I would be a little bit wary about this because this may run the autocomplete file, including the `@customElement` decorator. Now that I think about it, I think the reason we used `any` was to get around having to include the other components like this (not exactly a good reason, though).

That said, in this case, because `Autocomplete` is only being used as a type, TypeScript is smart enough to compile it out of the JavaScript bundle so that nothing from `@groupby/elements-autocomplete` is actually run. However, this does mean that we'll have to add `@groupby/elements-autocomplete` as a dev dependency when we get around to doing so.

In short, no action needed, but things to keep in mind.

packages/web-components/@groupby/elements-sayt/test/unit/sayt.test.ts
```diff
@@ -270,28 +270,30 @@ describe('Sayt Component', () => {
...
       expect(searchboxAddEventListener).to.be.calledWith('keydown', sayt.changeSelection);
+      expect(searchboxAddEventListener).to.be.calledWith('keydown', sayt.updateSearchboxInputTerm, true);
       expect(windowAddEventListener).to.not.be.calledWith(SEARCHBOX_INPUT);
     });
 
     it('should remove an event listener if provided a `remove` paramater and an input ID and it exists on the page', () => {
       const searchboxRemoveEventListener = spy();
       const windowRemoveEventListener = stub(window, 'removeEventListener');
-      const getElementById = stub(document, 'getElementById').returns({ removeEventListener: searchboxRemoveEventListener });
+      const getElementById = stub(document, 'getElementById').returns({ removeEventListener: searchboxRemoveEventListener, parentElement: { removeEventListener: searchboxRemoveEventListener } });
```
geoffcoutts: Same as above

packages/web-components/@groupby/elements-sayt/src/sayt.ts
```diff
@@ -225,12 +228,27 @@ export default class Sayt extends Base {
...
+   * Processes a keyboard event from the searchbox component's parent element, in order to update the search term
+   * in the input box.
+   * The `updateSearchTerm` method on autocomplete is called in response to the keyboard event key `enter`.
+   *
+   * @param event The KeyboardEvent object.
+   */
+  updateSearchboxInputTerm(event: KeyboardEvent): void {
+    if (event.key === 'Enter' && this.nodeInSearchbox(event.target as Node)) {
+      const autocomplete = this.querySelector<Autocomplete>('[data-gbe-ref="autocomplete"]');
+      autocomplete.updateSearchTerm(this.group);
```
geoffcoutts: Should this have a check for whether `autocomplete` exists?

packages/web-components/@groupby/elements-sayt/test/unit/sayt.test.ts
```diff
@@ -1085,4 +1087,41 @@ describe('Sayt Component', () => {
...
+  describe('updateSearchboxInputTerm()', () => {
+    let updateSearchTerm;
+    let group;
+
+    beforeEach(() => {
+      updateSearchTerm = spy();
+      group = sayt.group = 'group-1';
+      stub(sayt, 'querySelector')
+        .withArgs('[data-gbe-ref="autocomplete"]')
+        .returns({ updateSearchTerm, group });
```
geoffcoutts: `group` being on autocomplete isn't relevant to this function and can be removed from the stub.

packages/web-components/@groupby/elements-autocomplete/test/unit/autocomplete.test.ts
```diff
@@ -455,4 +487,21 @@ describe('Autcomplete Component', () => {
...
+      stub(autocomplete, 'selectedItem').get(() => ({ label: 'dress' }));
+      const group = autocomplete.group = 'group-1';
+      const payload = {
+        term: autocomplete.selectedItem.label,
+        group,
+        search: false,
+      };
+      const dispatchElementsEvent = stub(autocomplete, 'dispatchElementsEvent');
+
+      autocomplete.requestUpdateSearchTerm(group);
```
geoffcoutts: This does not take in `group` (anymore?). 
```suggestion
      autocomplete.requestUpdateSearchTerm();
```

packages/web-components/@groupby/elements-sayt/src/sayt.ts
geoffcoutts: Given that this is being introduced now, shouldn't `@groupby/elements-autocomplete` be added as a dev dependency to this package as part of this ticket?

packages/web-components/@groupby/elements-sayt/test/unit/sayt.test.ts
```diff
@@ -1085,4 +1095,50 @@ describe('Sayt Component', () => {
...
+
+    it('should not call the requestUpdateSearchTerm method on autocomplete if the event node is not a child of searchbox', () => {
+      stub(sayt, 'nodeInSearchbox').returns(false);
+
+      sayt.updateSearchboxInputTerm({ key: 'Enter' });
+
+      expect(requestUpdateSearchTerm).to.not.be.called;
+    });
+
+    it('should not attempt to call the requestUpdateSearch term method on autocomplete if autocomplete does not exist', () => {
```
geoffcoutts: This test also needs this stub I believe `stub(sayt, 'nodeInSearchbox').returns(true);`.

packages/web-components/@groupby/elements-sayt/src/sayt.ts
ChantalDesrochers: We haven't been adding our packages as dev dependencies. I assumed we were waiting to publish? Wouldn't ``@groupby/elements-base`` also be a dev dependency of sayt?

jmbtutor: `@groupby/elements-base` would be a real dependency of sayt. We haven't released anything yet, so we can push this off until we've published our packages (otherwise we don't have anything to point to)

demo/temp/gbe-all-plugins-bundle.js
```diff
@@ -10401,7 +10401,7 @@ function plural(ms, n, name) {
 /* WEBPACK VAR INJECTION */(function(global) {/*!
```
jmbtutor: Especially since we're so close to releasing things, we don't want to be introducing changes to this bundle. Please revert.

packages/web-components/@groupby/elements-sayt/src/sayt.ts
```diff
@@ -243,7 +243,7 @@ export default class Sayt extends Base {
    * @param event The KeyboardEvent object.
    */
   updateSearchboxInputTerm(event: KeyboardEvent): void {
-    if (event.key !== 'Enter' || !this.nodeInSearchbox(event.target as Node)) return;
+    if (event.key !== 'Enter' || !this.nodeInSearchbox(event.target as Node) || !this.visible) return;
```
jmbtutor: The least expensive operation should go first to take advantage of short-circuiting. I would first check `visible`, then `key`, then `nodeInSearchbox`. That way, the runtime doesn't have to check the DOM tree if it doesn't have to.

packages/web-components/@groupby/elements-sayt/test/unit/sayt.test.ts
```diff
@@ -1108,6 +1108,7 @@ describe('Sayt Component', () => {
 
     it('should call the requestUpdateSearchTerm method on autocomplete when enter is pressed and the event target is within the searchbox component', () => {
       stub(sayt, 'nodeInSearchbox').returns(true);
+      sayt.visible = true;
```
jmbtutor: Nitpick, but I would put assignments first before function calls. Same goes for the other instances introduced below.

packages/web-components/@groupby/elements-sayt/test/unit/sayt.test.ts
```diff
@@ -1132,6 +1135,20 @@ describe('Sayt Component', () => {
 
     it('should not attempt to call the requestUpdateSearch term method on autocomplete if autocomplete does not exist', () => {
       stub(sayt, 'nodeInSearchbox').returns(true);
+      sayt.visible = true;
+      sayt.querySelector.restore();
+      stub(sayt, 'querySelector')
```
jmbtutor: This can all be on one line if it's not too long.

packages/web-components/@groupby/elements-sayt/test/unit/sayt.test.ts
```diff
@@ -1085,4 +1095,64 @@ describe('Sayt Component', () => {
...
+    it('should not call the requestUpdateSearchTerm method on autocomplete if a key other than enter is pressed', () => {
+      stub(sayt, 'nodeInSearchbox').returns(true);
+      sayt.visible = true;
+
+      sayt.updateSearchboxInputTerm({ key: 'Z' });
+
+      expect(requestUpdateSearchTerm).to.not.be.called;
+    });
+
+    it('should not call the requestUpdateSearchTerm method on autocomplete if the event node is not a child of searchbox', () => {
```
jmbtutor: We want it to work when the target _is_ the searchbox, so saying it's a child doesn't work. Also, please try to be consistent in your wording between these related tests.

```suggestion
    it('should not call the requestUpdateSearchTerm method on autocomplete if the event target is not contained in searchbox', () => {
```

PR #62, opened 2019-11-29
ELE-530: Preparation v0.1.0 for manual release

ChantalDesrochers:


packages/web-components/@groupby/elements-sayt/package.json
```diff
@@ -22,13 +22,18 @@
     "url": "https://github.com/groupby/elements-view.git",
     "directory": "packages/web-components/@groupby/elements-sayt"
   },
-  "author": "",
+  "author": "Groupby Inc",
   "license": "MIT",
   "devDependencies": {
     "html-entities": "^1.2.1"
```
geoffcoutts: This should be moved to the `devDependencies` at the top level. Confirm with other packages as well, please.

packages/web-components/@groupby/elements-sayt/package.json
```diff
@@ -22,16 +22,18 @@
     "url": "https://github.com/groupby/elements-view.git",
     "directory": "packages/web-components/@groupby/elements-sayt"
   },
-  "author": "",
+  "author": "Groupby Inc",
   "license": "MIT",
-  "devDependencies": {
-    "html-entities": "^1.2.1"
+  "peerDependencies": {
+    "@groupby/elements-autocomplete": "^0.1.0"
```
DanielMoniz-GBI: I believe this should be a direct dependency. From the top of `sayt.ts`:
```typescript
import { Autocomplete } from '@groupby/elements-autocomplete';
```

Presumably Autocomplete could specify Sayt as a peer dependency (which it currently does not).

jmbtutor: After some in-person discussion, we concluded that autocomplete should be a dev dependency because it is only used at build time (no runtime objects are being imported, just an interface). At most, it can be an optional dependency since it is not necessary for autocomplete to be present to use sayt, but that is less appropriate than a dev dependency because no objects are imported and used at runtime.

> Presumably Autocomplete could specify Sayt as a peer dependency (which it currently does not).

This is not desirable because Autocomplete would then be assuming its usage and creates a circular dependency.

